PR #48, opened 2023-05-24
fix(input): not emit odsValueChange on init component

aesteves60:


packages/libraries/core/src/components/input/ods-input-controller.ts
```diff
@@ -89,18 +89,22 @@ export class OdsInputController extends OdsComponentController<OdsInput> {
...
   }
 
-  onValueChange(value: OdsInputValue, oldValue?: OdsInputValue) {
+  onValueChange(value: OdsInputValue, oldValue?: OdsInputValue): void {
     this.logger.debug(`[input=${this.component.value}]`, 'value changed', { value, oldValue });
+    this.assertValue(value);
+    this.component.emitChange(value, oldValue);
+  }
+
+  assertValue(value: OdsInputValue): void {
```
aesteves60, line 105: can't be private because of test :(

PR #58, opened 2023-05-29
Feat/pagination v2

dpellier:


astagnol (commented):
- Previous and Next buttons icons color is different from XD (wrong from Pagination v1). We should change 800 to 500: https://github.com/ovh/design-system/blob/master/packages/stencil/components/pagination/src/components/osds-pagination/styles/osds-pagination.color.scss#L8

- Maybe we should add in usage.mdx an example with totalItems attribute?

- Waiting for Design team validation (staging check)

aesteves60 (approved):

PR #59, opened 2023-05-30
Feat/select cdk

aesteves60:


ovh-cds:
CDS Report build#623.1 ✘  
* 
  * build ✘

packages/libraries/cdk/src/components/surface/ocdk-surface.ts
```diff
@@ -246,6 +246,15 @@ export class OcdkSurface extends HTMLElement {
     this.controller.setOriginCorner(originCorner);
   }
 
+  /**
+   * Check if a click is outside of the surface
+   * @param event - event reference
+   */
+  isClickOutsideSurface(event: Event): boolean {
```
astagnol, line 253: We need to discuss if that test should be done directly here, in the surface, or should be done in controller for every component using CDK

packages/stencil/components/select/src/components/osds-select/osds-select.tsx
```diff
@@ -270,32 +292,41 @@ export class OsdsSelect implements OdsSelect<OdsStencilMethods<OdsSelectMethods>
 
     return (
       <Host {...{
-        class: `${disabled ? 'disabled' : ''}${opened ? ' opened' : ''}${this.hasError() ? ' ods-error' : ''}`,
+        class: `${disabled ? 'disabled' : ''}${isSurfaceOpen ? ' opened' : ''}${this.hasError() ? ' ods-error' : ''}`,
```
astagnol, line 295: isSurfaceOpen, is it working? Could not find it in DOM for class. Besides, maybe it won't be necessary since ocdk-surface (its class) already manages this. We can discuss about it

astagnol (changes requested):
- when Select is open, we should be able to close it by clicking again on the trigger.

- it is a bit different from the XD. When Select is open, trigger border-radius (bottom) should be set to 0 and trigger background-color should be var(--ods-color-primary-200)

- in the jira ticket, it mentions an update for "handling keyboard navigation on focus". It should work as mentioned on [functional specs](https://confluence.ovhcloud.tools/display/ODS/Select+-+Design) (Accessibility)

- you'll need to add tests regarding this keyboard navigation in the controller (you can check Tooltip, Menu or Popover for example)

dpellier:
Clicking on the trigger when open, does not close the select, it was the previous behavior though

packages/stencil/components/select/src/components/osds-select/osds-select.e2e.ts
```diff
@@ -260,4 +261,64 @@ console.log('before clicks');
...
     });
   });
+
+  describe('method:handlerKeyPress', () => {
+    it('should open select when Enter', async () => {
+      await setup({ });
+      await page.waitForChanges();
+      await el.focus();
+      await page.keyboard.press('Enter');
+      expect(await el.getProperty('opened')).toBeTruthy();
```
dpellier, line 271: we should try to avoid `toBeTruthy` and `toBeFalsy` when we know the exact expected value cause it can pass with the wrong value
prefer `toBe('true')`

packages/libraries/core/src/components/select/select/ods-select-controller.ts
```diff
@@ -51,6 +52,7 @@ export class OdsSelectController extends OdsComponentController<OdsSelect> {
   openSurface(): void {
     if (this.component.surface) {
       this.component.surface.open();
+      this.component.opened = true;
```
dpellier, line 55: update controller test?

packages/libraries/core/src/components/select/select/ods-select-controller.ts
```diff
@@ -42,6 +42,7 @@ export class OdsSelectController extends OdsComponentController<OdsSelect> {
   closeSurface(): void {
     if (this.component.surface?.opened) {
       this.component.surface.close();
+      this.component.opened = false;
```
dpellier, line 45: update controller test?

packages/stencil/components/select/src/components/osds-select/osds-select.tsx
```diff
@@ -150,6 +150,65 @@ export class OsdsSelect implements OdsSelect<OdsStencilMethods<OdsSelectMethods>
...
+  handleKeyDown(ev: KeyboardEvent): void {
+    ev.stopPropagation();
+    const isEnter = ev.code.includes('Enter');
+    const isEscape = ev.code.includes('Escape');
+    const isArrowDownOrUp = ev.code === 'ArrowUp' || ev.code === 'ArrowDown';
+    if (!isEnter && !isArrowDownOrUp && !isEscape) {
+      return;
+    }
+    if (isEscape) {
+      this.controller.closeSurface();
```
dpellier, line 162: return to avoid passing through the code next

dpellier (changes requested):

ovh-cds:
CDS Report build#695.0 ✘  
* 
  * build ✘

packages/libraries/core/src/components/select/select/ods-select-controller.spec.ts
```diff
@@ -25,27 +25,27 @@ describe('spec:ods-select-controller', () => {
 
   describe('methods', () => {
```
dpellier, line 26: please add tests to cover the added key handler methods

dpellier (changes requested):

dpellier (changes requested):

packages/tools/storybook/stories/components/select/select.web-components.stories.ts
```diff
@@ -58,7 +58,7 @@ export default {
 
 /* Default */
 const TemplateDefault = (args: any) => html`
-  <osds-select ...=${getTagAttributes(args)}>
+  <osds-select style="width:20%" ...=${getTagAttributes(args)}>
```
dpellier, line 61: avoid those on stories
- it should display default rendering
- you cannot test flex attr anymore

dpellier:
@aesteves60 Regarding key navigation; it feels a bit weird:
- focus is lost after pressing enter on a value
- you cannot select the current value again
- pressing enter when the trigger element is focus remove the current value

dpellier (approved):

astagnol (approved):

PR #61, opened 2023-06-01
fix(link): set href & target optionnal

aesteves60:


astagnol (changes requested):
Nice!
Can you add a sentence in usage.mdx explaining to users that when they don't add href attribute, osds-link is button-like?

astagnol (approved):

PR #74, opened 2023-06-19
Feat/switch

aesteves60:


ovh-cds:
CDS Report build#722.0 ✘  
* 
  * build ✘

ovh-cds:
CDS Report build#724.0 ✘  
* 
  * build ✘

ovh-cds:
CDS Report build#725.0 ✘  
* 
  * build ✘

packages/stencil/components/switch/src/components/osds-switch/styles/osds-switch.mixins.scss
```diff
@@ -0,0 +1,14 @@
+// general mixins and functions for the component
+
+@mixin osds-switch-on-selected-host() {
+  :host(&) {
+    @content;
+  }
+}
+
+@mixin osds-switch-on-switch-items-child($property: '') {
```
dpellier, line 9: seems weird to style from the parent
each child should contains its own style regarding of its state, which is passed from the parent

packages/stencil/components/switch/src/index.html
```diff
@@ -0,0 +1,148 @@
...
+    <osds-switch-item value="1" id="1" checked>Item 1</osds-switch-item>
+    <osds-switch-item value="2" id="2">Item 2</osds-switch-item>
+    <osds-switch-item value="3" id="3">Item 3</osds-switch-item>
+  </osds-switch>
+
+  <osds-switch size="md" color="promotion" contrasted>
+    <osds-switch-item value="1" id="1" checked>Item 1</osds-switch-item>
+    <osds-switch-item value="2" id="2">Item 2</osds-switch-item>
+    <osds-switch-item value="3" id="3">Item 3</osds-switch-item>
+  </osds-switch>
```
dpellier, line 142: could you add a test with components as item slot ?
like:
```
<osds-switch-item value="1" id="1" checked>
    <osds-icon name=.... />
    <osds-text>Text</osds-text>
</osds-switch-item>
```

packages/libraries/core/src/components/switch/ods-switch-controller.ts
```diff
@@ -0,0 +1,76 @@
...
+    return switchItems.find((_, i) => i === currentIndex - 1);
+  }
+
+  findNextSwitchItem(): HtmlSwitchItem | undefined {
+    const switchItems = this.getSwitchItems();
+    const currentIndex = this.getActiveSwitchItemIndex(switchItems);
+    if (currentIndex === undefined || (currentIndex + 1) >= switchItems.length) {
+      return undefined;
+    }
+    return switchItems.find((_, i) => i === currentIndex + 1);
```
dpellier, line 69: `return switchItems[currentIndex + 1]`?

packages/libraries/core/src/components/switch/ods-switch-controller.ts
```diff
@@ -0,0 +1,76 @@
...
+    return switchItems.findIndex(switchItem => switchItem.id === activeElement?.id);
+  }
+
+  findPreviousSwitchItem(): HtmlSwitchItem | undefined {
+    const switchItems = this.getSwitchItems();
+    const currentIndex = this.getActiveSwitchItemIndex(switchItems);
+    if (currentIndex === undefined || (currentIndex - 1) < 0) {
+      return undefined;
+    }
+    return switchItems.find((_, i) => i === currentIndex - 1);
```
dpellier, line 60: `return switchItems[currentIndex + 1]`?

dpellier (changes requested):
Can you move the whole `switch-item` controller dir into the `switch` ones qnd split both in two sub dir?

Regarding switch-item, I find it weird to have the style managed by the parent, to be discussed, as do we need to enforce child slot in some case?

packages/libraries/core/src/components/switch/ods-switch-default-attributes.ts
```diff
@@ -0,0 +1,18 @@
...
+import { OdsSwitchAttributes } from './ods-switch-attributes';
+import { OdsSwitchSize } from './ods-switch-size';
+
+export const odsSwitchDefaultAttributesDoc  = {
+    /**
+     * indicates if the pagination is entirely disabled.
+     * it means no interactions (hover, click, focus, etc)
+     */
+    disabled: false,
+    name: '',
```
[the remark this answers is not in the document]
aesteves60: I rework the affectation for have an error unknown attributes

packages/stencil/components/switch/src/components/osds-switch-item/osds-switch-item.tsx
```diff
@@ -0,0 +1,75 @@
...
+  /** @see odsSwitchItemClick */
+  @Event() odsSwitchItemClick!: EventEmitter<{ value: string }>;
+
+  handlerOnClick(event?: MouseEvent): void {
+    event?.preventDefault();
+    this.odsSwitchItemClick.emit({ value: this.value});
+  }
+
+  handlerOnKeyDown(event: KeyboardEvent): void {
+    const isEnter = event.code.includes('Enter');
```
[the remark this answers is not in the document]
aesteves60: yes it's for the `NumpadEnter`. I don't know if this is expected

packages/libraries/core/src/components/switch/ods-switch-controller.ts
```diff
@@ -0,0 +1,76 @@
...
+      newCheckedSwitchItem?.classList.add('fadein-from-right');
+    } else {
+      selectedSwitchItem?.classList.add('fadeout-from-right');
+      newCheckedSwitchItem?.classList.add('fadein-from-left');
+    }
+    return { current: newCheckedSwitchItem, old: selectedSwitchItem };
+  }
+
+  private setClassSwitchItems(switchItems: HtmlSwitchItem[]) {
+    switchItems.forEach(switchItem => switchItem.setAttribute('class', 'hydrated'));
```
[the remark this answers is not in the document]
aesteves60: Yes, I thought not

packages/libraries/core/src/components/switch/ods-switch-controller.ts
```diff
@@ -0,0 +1,76 @@
...
+  }
+
+  findSelectedSwitchItem(): HtmlSwitchItem | undefined {
+    return this.getSwitchItems().find(switchItem => switchItem.checked);
+  }
+
+  private getActiveSwitchItemIndex(switchItems: HtmlSwitchItem[]): number | undefined {
+    const activeElement = document.activeElement;
+    if (activeElement?.localName !== this.SWITCH_ITEM_TAG) {
```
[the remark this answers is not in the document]
aesteves60: image
Yes it's not the same. the selected is the item at the left and the active is the right

packages/libraries/core/src/components/switch/ods-switch-controller.spec.ts
```diff
@@ -108,16 +108,16 @@ describe('spec:ods-switch-controller', () => {
       expect(nextSwitchItem).toBe(undefined);
     });
 
-    it('should not findSelectedSwitchItem', () => {
+    it('should not finCheckedSwitchItem', () => {
```
dpellier, line 111: typo `findCheckedSwitchItem`

dpellier:
When running `yarn start` on the component dir:
```
[ WARN  ]  @Method requires async: ./src/components/osds-switch-item/osds-switch-item.tsx:47:3
           External @Method() setFocus() must return a Promise. Consider prefixing the method with async, such as
           @Method() async setFocus().

     L46:  @Method()
     L47:  setFocus(): void {
     L48:    this.el.focus();

[19:52.2]  build finished in 7.25 s
```

packages/stencil/components/switch/src/components/osds-switch-item/osds-switch-item.tsx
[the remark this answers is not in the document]
dpellier: Probably not, all other keyboard nav seems to only manage actual Enter

dpellier:
Is it on purpose that the `contrasted` style only apply to the selected element ?
Are there no design rules for unchecked ones ?

dpellier:
Issue with double focus on each item, plus if you type "Enter" on the square focus, the animation is triggered but the selection does not change, see video

https://github.com/ovh/design-system/assets/1390934/04dd0b7f-12e7-49c5-8bc6-61c0e9389abb

dpellier:
Also transition effect is kind of weird when switching from one element to another that is not direct sibling

dpellier:
Issue also with the radio being smaller than the item
Capture d’écran 2023-07-03 à 10 46 44

packages/stencil/components/switch/src/components/osds-switch-item/osds-switch-item.scss
```diff
@@ -17,9 +17,14 @@
...
 
-    & osds-radio > * {
-        cursor: pointer;
-    }
+    & osds-radio {
+        & > * {
+            cursor: pointer;
+        }
+        &:focus-visible {
+            outline-color: var(--ods-color-gray-500);
```
dpellier, line 25: radio don't have a focus style already?

packages/libraries/core/src/components/switch/ods-switch-controller.ts
```diff
@@ -17,6 +17,7 @@ export class OdsSwitchController extends OdsComponentController<OdsSwitch> {
 
   changeCheckedSwitchItem(value: string): { current: HtmlSwitchItem, previous: HtmlSwitchItem } {
     const switchItems = this.getSwitchItems();
+    switchItems.forEach(switchItem => switchItem.className = 'hydrated');
```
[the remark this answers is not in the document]
dpellier: Seems because you're not removing previous transition class.
You should remove only the ones you managed and not reset the whole class list

dpellier (approved):

packages/themes/blue-jeans/theme/theme.scss
```diff
@@ -217,6 +217,16 @@
         lg: (width: calc(21.875rem * var(--ods-size-ratio))),
         xl: (width: calc(31.25rem * var(--ods-size-ratio))),
       ),
+      $switch: (
+        sm: (
+          height: var(--ods-size-8),
+          line-height: var(--ods-size-8),
+        ),
+        md: (
+          height: var(--ods-size-10),
```
astagnol, line 226: Update md height from 40px to 36px: height: `calc(2.25rem * var(--ods-size-ratio))`

packages/themes/blue-jeans/theme/theme.scss
```diff
@@ -217,6 +217,16 @@
         lg: (width: calc(21.875rem * var(--ods-size-ratio))),
         xl: (width: calc(31.25rem * var(--ods-size-ratio))),
       ),
+      $switch: (
+        sm: (
+          height: var(--ods-size-8),
```
astagnol, line 222: In browser, height is 21px instead of 24px. In Chromium, it is saying the variable is not defined

packages/stencil/components/switch/src/components/osds-switch/styles/osds-switch.typography.scss
```diff
@@ -0,0 +1,14 @@
+@import '~@ovhcloud/ods-theming/typography/ods-theming-typography';
+@import '~@ovhcloud/ods-theming/typography/ods-component-typography';
+@import './osds-switch.mixins';
+
+// Main CSS mixin for typography
+@mixin osds-switch-theme-typography() {
+  :host {
+    $typography-properties: ods-get-typography-properties(body, '400');
```
astagnol, line 8: According to XD sketches, we should be using subheading '100' instead of body '400': 16px, SemiBold, line-height of 20px.
Still, currently it does the job since there's a font-weight: 600 on osds-radio so...

astagnol (changes requested):
- Switch item goes on next line when there is not enough place so a `white-space: nowrap` should be added 

- we should have color p-100 (similar to hover) for active item by keyboard navigation

- when Switch is disabled, we can still focus by keyboard navigation and I think an event.preventDefault() might be missing since event is still triggered. It causes trouble when you go back to 'not disabled' state

astagnol (approved):

astagnol:
Should be merged after Radio fix (focus + event still triggered when disabled)

PR #93, opened 2023-07-04
feat(pagination): add tooltip on arrow

aesteves60:


astagnol (changes requested):
Capture d’écran 2023-07-06 à 10 54 57
Tooltip surface max-width used in Pagination should be 300px

In Storybook, we're losing focus when typing in labelTooltipPrevious & labelTooltipNext fields

packages/stencil/components/pagination/src/components/osds-pagination/osds-pagination.tsx
```diff
@@ -172,36 +177,43 @@ export class OsdsPagination implements OdsPagination<OdsStencilMethods<OdsPagina
...
-              this.handleNextClick(Number(this.current));
-            }
-          }}
-          size={OdsButtonSize.sm}>
-          <osds-icon
-            size={OdsIconSize.sm}
-            name={arrowIcon}
+        <osds-tooltip>
+          <osds-tooltip-content slot="tooltip-content">
+            { isLeft ? this.labelTooltipPrevious : this.labelTooltipNext }
```
dpellier, line 187: did you try using `key` instead to trigger rerender on change?
`forceUpdate` on the whole element is quite costly, would be nice to avoid using it

dpellier (approved):

packages/libraries/cdk/src/components/surface/strategies/symmetry/ocdk-surface-symmetry.bl-tl.ts
```diff
@@ -38,7 +38,7 @@ export function ocdkSurfaceSymmetryBlTl(): OcdkSurfaceOnePositionStrategy<OcdkSu
         },
         appliers: {
           maxHeight: (opt) => opt.inspections.comfort.availableBottom,
-          maxWidth: (opt) => opt.measurements.surfaceSize.width,
+          maxWidth: (opt) => opt.inspections.comfort.availableLeft,
```
astagnol, line 41: Should be availableRight instead of left

astagnol (changes requested):

astagnol (approved):

PR #98, opened 2023-07-10
fix(select): multiples accessibility fixes

skhamvon:


packages/stencil/components/select/src/components/osds-select/osds-select.tsx
```diff
@@ -253,6 +253,13 @@ export class OsdsSelect implements OdsSelect<OdsStencilMethods<OdsSelectMethods>
     }
     this.logger.log('[checkForClickOutside]', arguments, { validity: this.validityState });
     this.controller.closeSurface();
+
+    const selectOptions = this.el.querySelectorAll('osds-select-option');
```
dpellier, line 257: can be done in a single loop to increase perf, something like :
```
this.el.querySelectorAll('osds-select-option').forEach((option) => {
  if (this.value && option.getAttribute('value') === this.value) {
    option.setAttribute('selected', 'true');
  } else {
    option.setAttribute('selected', 'false');
  }
})
```

dpellier (changes requested):

packages/stencil/components/select/src/components/osds-select/osds-select.tsx
aesteves60: You can also get the list of osds-select-options un the controller with `this.controller.selectOptions`

skhamvon: Fixed !

feoche:
Changing scope & behavior regarding Select usage

PR #101, opened 2023-07-17
fix(input): set cursor to not-allowed when disabled

Leotheluck:
Component input should have cursor to not-allowed when disabled, and pointer-events shouldn't be accessible (in this case, the native input has disabled attribute which disables native pointer-events, prioritizing cursor in the styling)


dpellier (changes requested):
You have included generated file from your local remains of form-field.
Please clean up by deleting all the form-field related files on this branch, then rebase again on master, only the osds-input.scss file change should remain.

packages/stencil/components/input/src/components/osds-input/osds-input.scss
```diff
@@ -49,14 +58,6 @@ input[type=number]::-webkit-inner-spin-button {
   margin: 0;
 }
 
-[disabled] {
-  @include osds-input-on-selected-host() {
-    opacity: .5;
-    cursor: not-allowed;
-    pointer-events: none;
```
dpellier, line 56: is it what has been asked for? cause this now allows user to select content on disabled input which was not doable before

Leotheluck: goal was to display cursor as not allowed but not be able to interact with the input (which is enabled by the native disabled attribute of it)

dpellier (approved):

aesteves60 (approved):

PR #113, opened 2023-07-26
fix(input): block actions if disabled

Leotheluck:
Inputs with clearable or passwords with masked were interactive if the input was disabled. This fix covers both cases, adds a specific cursor to the icons if the input is disabled and adds an e2e test to check if it works as intended.


dpellier (changes requested):
Capture d’écran 2023-07-26 à 15 32 53

`masked` seems to be broken on storybook

astagnol (approved):

aesteves60:
Miss a test for the `hide` and spec test

dpellier (approved):

aesteves60 (approved):

PR #116, opened 2023-07-27
fix(select): unselect non validated option after closing

skhamvon:


dpellier (approved):

astagnol (changes requested):
Open Select => go to value 3 via Tab => escape => re-open Select => Tab => OK (back to value 1)

But open Select => got to value 3 via Tab => click outside the Select => re-open Select => Tab => not correct, back to value 2 instead

astagnol (approved):

PR #118, opened 2023-07-28
fix(select): navigation with keys improvments

skhamvon:


astagnol (commented):
It could be great to check if both issues regarding Select accessibility are working as expected together (loop + unselect non validated option after closing)

astagnol (approved):

PR #125, opened 2023-08-01
docs(storybook): add versions selector in sidebar

feoche:


packages/tools/storybook/.storybook/manager.js
```diff
@@ -2,6 +2,45 @@ import { addons } from '@storybook/addons';
...
+    }
+  }, 500);
+};
+
+setTimeout(() => {
+  const urlVersionRegex = /\/(latest|v(\d\.?){3}[^\/]*)\//gmi;
+  if (urlVersionRegex.test(location.href)) {
+    const releasesSelector = document.createElement('select');
+    releasesSelector.id = 'release-selector';
+    Object.entries(releases).map((value) => {
```
dpellier, line 24: prefer `forEach` if you don't care about what is returned

packages/tools/storybook/.storybook/manager.js
```diff
@@ -2,6 +2,45 @@ import { addons } from '@storybook/addons';
...
+      const version = selection.target.value === 'latest' ? 'latest' : `v${selection.target.value}`;
+      location.pathname = location.pathname.replace(urlVersionRegex, version);
+    };
+  } else {
+    const version = document.createElement('div');
+    version.id = 'release-selector';
+    version.innerText = Object.keys(releases)[0];
+    createHeaderElement(version);
+  }
+}, 2000);
```
dpellier, line 42: why do you need to wait here?

packages/tools/storybook/.storybook/manager.js
```diff
@@ -2,6 +2,45 @@ import { addons } from '@storybook/addons';
 import theme from './ods.theme';
 import React from 'react';
 
+import { releases } from '../public/releases.js';
+
+const createHeaderElement = (element) => {
+  const timer = setInterval(function(){
```
dpellier, line 8: 😧 is this really necessary?
can't you just add a dom loaded listener to start all your changes instead?

dpellier (changes requested):
(careful with your branch name, it should relates to the work done)

I don't see the version select on the deployed doc, is it normal?

feoche:
Outdated by #133

PR #138, opened 2023-08-08
fix(code): display error when inserting long code inputs

feoche:


packages/stencil/components/code/src/components/osds-code/osds-code.scss
```diff
@@ -9,24 +9,34 @@
...
-    justify-content: space-between;
+    display: grid;
+    grid-template-columns: 1fr auto;
+    grid-template-rows: 1fr;
+    gap: 0;
+    grid-template-areas:
+    ". .";
+    justify-content: start;
+    align-content: start;
+    grid-auto-columns: 100%;
```
dpellier, line 20: Could be a nice opportunity to normalize and simplify its style

you don't have to put all thos property, container would actually only needs:
```
display: grid;
grid-template-columns: 1fr auto;
grid-column-gap: var(--ods-size-code-md-padding);
```
the only issue is that this will add a gap even if code is not display, so `flex` is easier here:
```
display: flex;
flex-flow: row;
grid-column-gap: var(--ods-size-code-md-padding);
```
(and remove `width: 100%;`)

Then remove padding from both children.

This way padding is consistent across the whole component without defining any on child elements;
WDYT?

dpellier (changes requested):

dpellier (changes requested):
still an issue with the button hover triggering on the whole height

https://github.com/ovh/design-system/assets/1390934/52127132-6551-4a34-9aa8-fe61d4eac97b

PR #171, opened 2023-08-28
Refactor/file

aesteves60:
No Breaking because it's a new component


ovh-cds:
CDS Report build#1304.0 ✘  
* 
  * build ✘

packages-new/components/file/src/components/osds-file-header/interfaces/attributes.ts
```diff
@@ -0,0 +1,15 @@
+interface OdsFileHeaderAttribute {
+  acceptedTypes?: string;
+  dividerLabel?: string;
+  disabled?: boolean;
```
dpellier, line 4: almost in order ;)

packages-new/components/file/src/components/osds-file-item/core/controller.spec.ts
```diff
@@ -0,0 +1,52 @@
+import { OdsFileItemAttribute } from '../interfaces/attributes';
```
dpellier, line 1: type

packages-new/components/file/src/components/osds-file-item/core/controller.ts
```diff
@@ -0,0 +1,26 @@
...
+  validateAttributes(): void {
+    if (!this.component.name) {
+      throw new Error('ods-file-item: name attribute is required');
+    }
+    if (!this.component.size) {
+      throw new Error('ods-file-item: size attribute is required');
+    }
+  }
+  getSizeInKiloBytes(size: number) {
+    const result = Math.round(size / 1000);
```
dpellier, line 18: Seems a bit too simple to render human readable size, especially on large files, you could use instead a formatter function like:
```
const byteSizes = ['B', 'KB', 'MB', 'GB', 'TB']; // put this on top outside of controller class

// controller method
function formatBytes(bytes, decimals = 0) {
  const k = 1024;
  const dm = decimals < 0 ? 0 : decimals:
  const i = Math.floor(Math.log(bytes) / Math.log(k));

  return `${parseFloat((bytes / (k ** i)).toFixed(dm))} {byteSizes[i]}`;
}
```

feoche:
- [ ] Should not have `cursor: pointer` on not-clickable areas
- [ ] Should have `cursor: move` when drag-and-drop 
- [ ] Inner link has error outline when File is in error state
- [ ] Storybook : when changing fields after File is in error state, component is emptied 
image
- [ ] ^ Same with removing all files
- [ ] Should have min-width (344px)
- [ ] Storybook : not able to see File success state
- [ ] Storybook : Spec file does not work
- [ ] Max files alert should be displayed after uploading files (ex: with 4 files, add 2 files - 5th should be displayed, then after display error message)
- [ ] a11y : no action on Button/Links with Enter/Space

packages-new/components/file/src/components/osds-file-item/osds-file-item.tsx
```diff
@@ -0,0 +1,68 @@
...
+  }
+  getContent() {
+    const normalContent = <div class='ods-file__item'>
+    <osds-icon color={OdsThemeColorIntent.primary} name={ODS_ICON_NAME.FILE} size={ODS_ICON_SIZE.sm}></osds-icon>
+      <div class={{
+        'ods-file__item__label': true,
+        'ods-file__item__label--errored': this.error ?? false,
+      }}>
+        <span>{this.name}</span>
+        <span> ({this.getSizeInKiloBytes(this.size)} Ko)</span>
```
dpellier, line 40: remove Ko and prefer a formatter function (see controller review)

packages-new/components/file/src/components/osds-file/osds-file.scss
```diff
@@ -0,0 +1,178 @@
...
+// CSS targeted only for this component
+// /!\ for theming purposes, it has to be done in theming files
+// (i.e. packages/libraries/theming/...)
+:host {
+
+}
+
+:host([attribute]) {
+
+}
```
dpellier, line 17: I guess you can remove the empty hosts

packages-new/components/file/src/components/osds-file/osds-file.scss
```diff
@@ -0,0 +1,178 @@
+@import '~@ovhcloud/ods-theming/color/ods-theming-color';
+@import '~@ovhcloud/ods-theming/ods-theme';
+@import 'styles/osds-file.mixins';
+@import 'styles/osds-file.color';
+@import 'styles/osds-file.size';
+@import 'styles/osds-file.typography';
```
dpellier, line 6: you can remove those

packages-new/components/file/src/components/osds-file/public-api.ts
```diff
@@ -0,0 +1 @@
+export { OsdsFile } from './osds-file'
```
dpellier, line 1: export also interfaces (attr and events)

packages-new/components/file/stencil.config.ts
```diff
@@ -0,0 +1,29 @@
...
+const args = process.argv.slice(2);
+
+export const config: Config = getStencilConfig({
+  namespace: 'osds-file',
+  args,
+  jestConfig: jestConfig.default,
+  reactOutput: {
+    componentCorePackage: '@ovhcloud/ods-component-file',
+    // exclude peer dependencies that corresponds to www usage
+    excludeComponents: ['osds-icon', 'osds-link', 'osds-text']
```
dpellier, line 14: regarding component dependencies, should be
`['osds-button', 'osds-icon', 'osds-link', 'osds-progress-bar']`

dpellier (changes requested):

PR #182, opened 2023-08-31
feat(phone-number): add new iso code & file trad for countries

aesteves60:
- Add new iso code
- file traduction for countries (en & fr)
- Flag ko pour unia
image-2023-08-31-14-48-09-779


aesteves60:
Iso code in the ODS enum but not in the trad file
 [ 'eu', 'un', 'unia', 'xk' ]
 Should I remove it or add in the trad file but when we create all the other file we need to add it too

dpellier:
> Iso code in the ODS enum but not in the trad file [ 'eu', 'un', 'unia', 'xk' ] Should I remove it or add in the trad file but when we create all the other file we need to add it too

`eu`, `un` and `unia` are not specific country, so they should not appear on the country code list.
But, they do have flags, so they should appear in the flag list, so extend the country list to create a flag list with those.

For `xk` I think we can remove it as it's not part of the official list: https://fr.wikipedia.org/wiki/Domaine_national_de_premier_niveau_du_Kosovo

packages-new/components/flag/src/components/osds-flag/constants/flag-iso-code.ts
```diff
@@ -0,0 +1,261 @@
+enum ODS_FLAG_ISO_CODE {
```
dpellier, line 1: I know we can't extend enum, but we shouldn't duplicate all the codes here.
You can use a normal JS object in this case, something like:
```
// your iso list from core
enum ISO {
  en = 'EN',
  fr = 'FR',
}

// your flag constant code
const FLAG = {
  ...ISO,
  EU: 'EU',
}

console.log(ISO) // Object { en: "EN", fr: "FR" }
console.log(FLAG) // Object { en: "EN", fr: "FR", EU: "EU" }
```

dpellier (changes requested):

packages-new/components/flag/src/components/osds-flag/constants/flag-iso-code.ts
aesteves60: I did it, but I'm not convinced by the lost of the typage and we lost the type of ODS_COUNTRY_ISO_CODE in the phone-number too

packages-new/components/flag/src/components/osds-flag/constants/flag-iso-code.ts
```diff
@@ -0,0 +1,18 @@
+import { ODS_COUNTRY_ISO_CODE } from "@ovhcloud/ods-common-core";
+
+const ODS_FLAG_ISO_CODE = {
+    ...ODS_COUNTRY_ISO_CODE,   
+    EU: 'eu', // Europe
+    UN: 'un', // United Nations
+    XK: 'xk', // Kosovo
+} as const;
```
aesteves60, line 8: Object.freeze break the typage

dpellier: For history purpose after solving it together: issue is with typescript version < 4.7.*
Solution is to enforce IDE to use the workspace version

PR #183, opened 2023-09-01
fix(flag): all flags storybook page

skhamvon:
ODS-872
Fix the path to load the flag correctly on the all flags storybook page


dpellier:
I will put the fix on the moved project on the refactor/storybook branch

dpellier:
Fix applied to the new storybook project 👍

PR #184, opened 2023-09-01
fix(radio-group): fix undefined error onDestroy

skhamvon:
Fix undefined error in console for radio-group 
ODS-873


dpellier (commented):
@skhamvon were you able to investigate why the `radioGroup` does not have an `el` although it should be part of the component?

dpellier (approved):

PR #189, opened 2023-09-06
Feat/phone number component with countries, locale & isoCode

aesteves60:
Add all countries, a specific list of countries, a select component for choice the current countrie.
Locale for the language. Can be auto calculate with browser preference.
IsoCode for the selected value in the select. Can be auto calculate with browser preference.

image


packages/components/phone-number/src/components/osds-phone-number/core/controller.ts
```diff
@@ -0,0 +1,56 @@
...
+    }
+    return this.component.countries || [];
+  }
+
+  private getNavigatorLang(): ODS_COUNTRY_ISO_CODE | undefined {
+    const language = navigator.language?.split('-')[0];
+    if (language && this.isOdsCountryCode(language)) {
+      return language;
+    }
+    const language2 = navigator.language?.split('-')[1];
```
aesteves60, line 42: Why ? Because navigator.language is a string like `en-us`, `fr-fr` but `fr` it's an isoCode 👍 but `en` no 
And we have many language like `fr-be`, `fr-ca`, `en-uk`, `en-us`...

dpellier: Only first part of the the code should be used, the second part is to specify dialect but it's not meant to be an actual locale.
If the first part is not managed in our list, just directly fallback to default locale

aesteves60: Yes but all the english country will not work correctly

packages/components/phone-number/src/components/osds-phone-number/osds-phone-number.tsx
```diff
@@ -1,33 +1,86 @@
...
+        {
+          this.hasCountries && <osds-select
+            tabindex="0"
+            disabled={this.disabled}
+            error={this.error}
+            value={this.isoCode}>
+            { this.countriesList?.map((country) => <osds-select-option value={ country }>
+              <div class="phone-number--select-option">
+                <osds-flag iso={country}></osds-flag>
+                <span>{ this.i18nCountries?.get(country)?.name }</span>
```
dpellier, line 74: osds-text?

packages/components/phone-number/src/components/osds-phone-number/osds-phone-number.scss
```diff
@@ -1,3 +1,36 @@
 @import '~@ovhcloud/ods-common-theming/color/ods-theming-color';
 @import '~@ovhcloud/ods-common-theming/ods-theme';
 
+:host {
+  display: inline-flex;
+
+  osds-select {
```
dpellier, line 7: if we go with BEM, we should set class with correct type (element, modifier) on every element and never use generic selector.
Here your code should look like a representation of your DOM, something more like:

```
.phone-number {
  ...

  &__select {
    ...

    &__option {
      ...

      &__flag {
        ...
      }
    }
  }

  &__input {
    ...

    &--first {
      ...
    }
  }
}
```

packages/components/phone-number/src/components/osds-phone-number/core/controller.ts
dpellier: Issue is there is a difference between `isoCode` that we use for flag => indeed we need the second part
And the `locale` which is always `fr`, `en`.
But this is used on both function

dpellier (changes requested):

packages/components/phone-number/src/components/osds-phone-number/constants/phone-number-countries.ts
```diff
@@ -0,0 +1,7 @@
+enum ODS_PHONE_NUMBER_COUTRIE {
```
dpellier, line 1: `ODS_PHONE_NUMBER_COUNTRY`
maybe `ODS_PHONE_NUMBER_COUNTRY_PRESET` would be even better

packages/components/phone-number/src/components/osds-phone-number/osds-phone-number.scss
```diff
@@ -1,3 +1,35 @@
 @import '~@ovhcloud/ods-common-theming/color/ods-theming-color';
 @import '~@ovhcloud/ods-common-theming/ods-theme';
 
+:host(.phone-number) {
```
dpellier, line 4: is the host declaration necessary?
can't you put this directly under `.phone-number`?

packages/components/phone-number/src/components/osds-phone-number/osds-phone-number.tsx
```diff
@@ -1,33 +1,98 @@
...
 
 /**
  * @slot (unnamed) - Phone Number content
  */
 @Component({
   tag: 'osds-phone-number',
   styleUrl: 'osds-phone-number.scss',
-  shadow: true
+  shadow: true,
+  assetsDirs: ['../../assets'],
```
dpellier, line 17: remove

packages/components/phone-number/src/components/osds-phone-number/osds-phone-number.tsx
```diff
@@ -1,33 +1,98 @@
...
+    const translationFile = this.controller.loadTranslationFileByLocale(locale);
+    this.i18nCountriesMap = translationFile.reduce((acc, country) => {
+      acc.set(country.isoCode, country);
+      return acc;
+    }, new Map());
+  }
+
+  @Watch('countries')
+  handlerCountries(): void {
+    this.hasCountries = Boolean(this.countries?.length);
```
dpellier, line 66: did we go with Boolean instead of `!!`, not sure anymore 😅

dpellier (changes requested):

packages/components/phone-number/src/components/osds-phone-number/osds-phone-number.tsx
aesteves60: We say both but I can replace the Boolean

packages/components/phone-number/src/components/osds-phone-number/osds-phone-number.scss
aesteves60: No It's not working just un the `.phone-number` 😢

dpellier:
@aesteves60 would be nice also to parse the countries attr on load to allow this kind of syntax:
```
<osds-phone-number countries="['fr','gb']" id="phoneNumberWithSetCountries" clearable value="+33 6426658925"></osds-phone-number>
```
WDYT?
(you can check breadcrumb to see an example)

aesteves60:
> @aesteves60 would be nice also to parse the countries attr on load to allow this kind of syntax:
> 
> ```
> <osds-phone-number countries="['fr','gb']" id="phoneNumberWithSetCountries" clearable value="+33 6426658925"></osds-phone-number>
> ```
> 
> WDYT? (you can check breadcrumb to see an example)

We should normalize this for all the component and for now all the new component with an array ! 
@dpellier Can I create a helper for this ?

packages/components/phone-number/src/components/osds-phone-number/interfaces/attributes.ts
```diff
@@ -1,11 +1,11 @@
 import type { ODS_COUNTRY_ISO_CODE, ODS_LOCALE } from "@ovhcloud/ods-common-core";
-import { ODS_PHONE_NUMBER_COUTRIE } from "../constants/phone-number-countries";
+import { ODS_PHONE_NUMBER_COUNTRY_PRESET } from "../constants/phone-number-countries";
 
 interface OdsPhoneNumberAttribute {
   /** Ability to clear the phone number value */
   clearable?: boolean;
   /** A specific subset of countries to display in the select instead of the whole list  */
-  countries?: ODS_COUNTRY_ISO_CODE[] | ODS_PHONE_NUMBER_COUTRIE.All;
+  countries?: ODS_COUNTRY_ISO_CODE[] | ODS_PHONE_NUMBER_COUNTRY_PRESET.All | string;
```
dpellier, line 8: should be `countries?: ODS_COUNTRY_ISO_CODE[] | ODS_PHONE_NUMBER_COUNTRY_PRESET| string;`
Even if we have only All for now, it should accept any value of it

dpellier (changes requested):

PR #202, opened 2023-09-13
feat(phone-number) Step 4 number placeholder & validation

aesteves60:
Add validation & placeholder with `google-libphonenumber`


packages/components/phone-number/src/components/osds-phone-number/interfaces/events.ts
```diff
@@ -0,0 +1,17 @@
...
+}
+
+interface OdsPhoneNumberEvent {
+  /** Send event with the input & the selected isoCode when le select value or the input value change */
+  odsValueChange: EventEmitter<OdsInputPhoneNumberValueChangeEventDetail>,
+}
+
+export {
+  OdsPhoneNumberEvent,
+  OdsInputPhoneNumberValueChangeEventDetail
```
dpellier, line 16: we should normalize the naming to be either `OdsPhoneNumber` or `OdsInputPhoneNumber`

packages/components/phone-number/src/components/osds-phone-number/osds-phone-number.tsx
```diff
@@ -51,35 +61,81 @@ export class OsdsPhoneNumber implements OdsPhoneNumberAttribute {
...
+    this.odsValueChange.emit({
+      ...event,
+      value: number && this.phoneUtil.format(number, PhoneNumberFormat.E164),
+      oldValue: oldNumber && this.phoneUtil.format(oldNumber, PhoneNumberFormat.E164) || undefined,
+      isoCode: this.isoCode,
+    })
+  }
+
+  @Watch('value')
+  private isValidValue(value: string | null): boolean {
```
dpellier, line 118: would be nice to split this between:
- the validity check, who return a boolean
- the error attribute update

otherwise code like that starts to be confusing:
```
if (this.isValidValue(this.value)) {
  // here we don't expect to have updated any component attribute
}
```

dpellier (changes requested):

astagnol (changes requested):
Capture d’écran 2023-09-14 à 11 24 09
* Countries + indicators in surface should be osds-text color 'text', level 'body' and size '400'
* osds-flag in surface should be 24x15px (and it will be that size too in Select when a country is selected)

dpellier (changes requested):
Seems like there is a perf issue on component state change.
Go to the "All countries" storybook page and toggle back and forth the error or disable toggle.
The component freeze.

ovh-cds:
CDS Report build#1471.0 ✘  
* 
  * build ✘

dpellier (approved):

astagnol (approved):

PR #207, opened 2023-09-20
feat(datepicker): step 2, styling the component

Leotheluck:
Datepicker component styling

2 new attributes:
> color?: defines the color of the Datepicker
> inline?: defines if the component should be inline

The datepicker module is displayed seamlessly border-to-border with the input above, input's bottom right corner radius is adjusted regarding the inline attribute.


dpellier (changes requested):
Can you update the doc to prevent storybook shortcut while typing in the input, as for now if you try to manually type the date, storybook will react when you type `/`

PR #211, opened 2023-09-21
fix(code): contrasted versions colors for "primary", "info", "text"

manoncarbonnel:
Fix for code component:
⚠️  Contrasted versions does not have the right colors for "primary", "info", "text"
👉  Fixed by updating the color palette


dpellier (approved):

astagnol (changes requested):
Text background-color variable is not set:
Capture d’écran 2023-09-21 à 17 38 29
Capture d’écran 2023-09-21 à 17 37 01

In ods-palette-definitions.scss, you should add '900-contrasted' (and '900') to 'blue-gray'. Then, it should be OK

astagnol (approved):

PR #215, opened 2023-09-26
fix(phone-number): style select + clear input error + sort list select options

aesteves60:
- Style
- select sort
- filter country not supported by the `google-libphonenumber`

image

### DIFF
--- a/packages/components/phone-number/src/components/osds-phone-number/osds-phone-number.scss
+++ b/packages/components/phone-number/src/components/osds-phone-number/osds-phone-number.scss
@@ -6,13 +6,14 @@
 }
 
 $phone-number-input-width: 236px;
+
 .phone-number {
 
   &__select {
     --ods-size-select-border-radius-bottom-right: 0;
     --ods-size-select-border-radius-top-right: 0;
     --ods-size-select-border-right-width: 0;
-    --ods-size-select-surface-width: calc(#{$phone-number-input-width} + 71px);
+    --ods-size-select-surface-width: #{$phone-number-input-width + 71px};
 
     &.opened + .phone-number__input {
       border-bottom-right-radius: 0;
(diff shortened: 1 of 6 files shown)


packages/components/phone-number/src/components/osds-phone-number/osds-phone-number.scss
```diff
@@ -5,12 +5,19 @@
...
 }
 
+$phone-number-input-width: 236px;
 .phone-number {
 
   &__select {
     --ods-size-select-border-radius-bottom-right: 0;
     --ods-size-select-border-radius-top-right: 0;
     --ods-size-select-border-right-width: 0;
+    --ods-size-select-surface-width: calc(#{$phone-number-input-width} + 71px);
```
dpellier, line 15: those are 2 hardcoded value, no need for `calc`

packages/components/phone-number/src/components/osds-phone-number/osds-phone-number.tsx
```diff
@@ -175,7 +185,7 @@ export class OsdsPhoneNumber implements OdsPhoneNumberAttribute, OdsPhoneNumberE
             }}
               slot="selectedLabel"
               class="phone-number__select-label" />
-            { this.parsedCountries?.map((country) => {
+            { this.parsedCountries?.sort((a, b) => this.sortCountriesByName(a, b)).map((country) => {
```
dpellier, line 188: using the sort directly in the render will go through it on every render, would be better to sort on Watch handler to ensure we sort only once

dpellier (changes requested):

manoncarbonnel (approved):

dpellier (approved):

astagnol (approved):

PR #219, opened 2023-09-28
feat(datepicker): step 3, excluding dates

Leotheluck:
Adding options to disable specific dates of the Datepicker.

New enum: 
`ODS_DATEPICKER_DAY`: monday = 1, saturday = 6, etc.

New attributes:
> datesDisabled?: Date[] (specific dates to disable)
> daysOfWeekDisabled?: ODS_DATEPICKER_DAY[] (ex: monday, sunday disabled for all dates)
> maxDate?: Date (maximum date selectable)
> minDate?: Date (minimum date selectable)

### DIFF
--- a/packages/components/datepicker/src/components/osds-datepicker/core/controller.ts
+++ b/packages/components/datepicker/src/components/osds-datepicker/core/controller.ts
@@ -23,10 +23,17 @@ class OdsDatepickerController {
       if (newValue === undefined || newValue === null || isNaN(newValue.getTime())) {
         this.component.value = null;
         this.component.datepickerInstance?.setDate({ clear: true });
+        this.component.emitDatepickerValueChange(null, oldValue);
       } else {
-        this.component.value = newValue;
-        this.component.datepickerInstance?.setDate(newValue);
-        this.component.emitValueChange(newValue, oldValue);
+        if (this.validateValue(newValue)) {
+          this.component.value = newValue;
+          this.component.datepickerInstance?.setDate(newValue);
+          this.component.emitDatepickerValueChange(newValue, oldValue);
+        } else {
+          this.component.value = null;
+          this.component.datepickerInstance?.setDate({ clear: true });
+          this.component.emitDatepickerValueChange(null, oldValue);
+        }
       }
     }
   }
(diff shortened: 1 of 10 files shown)


packages/components/datepicker/src/components/osds-datepicker/core/controller.ts
```diff
@@ -23,10 +23,17 @@ class OdsDatepickerController {
       if (newValue === undefined || newValue === null || isNaN(newValue.getTime())) {
```
dpellier, line 23: you can add also `|| !this.validateValue(newValue)` condition and remove the last else

packages/components/datepicker/src/components/osds-datepicker/core/controller.ts
```diff
@@ -35,6 +42,26 @@ class OdsDatepickerController {
...
     this.component.emitBlur();
   }
+
+  private getMidnightDate(date: Date) {
+    return new Date(Date.UTC(date.getFullYear(), date.getMonth(), date.getDate()))
+  }
+
+  private validateValue(value?: Date) {
+    if (value) {
+      if (!this.component.maxDate || this.getMidnightDate(value).getTime() <= this.getMidnightDate(this.component.maxDate).getTime() ) {
```
dpellier, line 52: you should split your validation and avoid calling the same function (`getMidnightDate`) so many times, more like

```
if (!value) {
  return false;
}

const midnightDate = new Date(Date.UTC(date.getFullYear(), date.getMonth(), date.getDate()))

if (this.component.maxDate) {
  return <test max date>
}

if (this.component.minDate) {
  return <test max date>
}

...
```

dpellier (changes requested):

packages/components/datepicker/src/components/osds-datepicker/core/controller.ts
aesteves60: Can you create a function for add context like `isValidValue`

packages/storybook/stories/components/datepicker/datepicker.web-components.stories.ts
```diff
@@ -38,6 +44,16 @@ const storyParams = {
     category: 'General',
     defaultValue: false,
   },
+  maxDate: {
+    category: 'General',
+    defaultValue: null,
+    options: [null, new Date('1999-11-02'), new Date('2024-11-02')],
+  },
+  minDate: {
```
aesteves60, line 52: I have an error on storybook with the date
image
And when I select one the input not display the value

aesteves60 (changes requested):

packages/components/datepicker/src/components/osds-datepicker/osds-datepicker.tsx
```diff
@@ -66,9 +79,32 @@ export class OsdsDatepicker implements OdsDatepickerAttribute, OdsDatepickerEven
...
+    }
+  }
+
+  @Watch('datesDisabled')
+  @Watch('daysOfWeekDisabled')
+  @Watch('format')
+  @Watch('maxDate')
+  @Watch('minDate')
+  updateDatepicker() {
+    if (this.datepickerInstance) {
```
aesteves60, line 94: Early return for less tabulation

aesteves60 (changes requested):

astagnol (changes requested):
Capture d’écran 2023-09-28 à 11 56 34

In Storybook, it can be a bit confusing for users when they choose a min and/or max date and the selected value is not displayed. Still, it's working as expected but I think we should display the selected value in select.

astagnol (approved):

aesteves60 (approved):

PR #221, opened 2023-09-28
fix(modal): closed in Storybook on window location change

manoncarbonnel:
:bug: In storybook, modal component keeps opened on route change and modals can overlap each other.

* Added a button to toggle modal in Storybook 'Docs' tab for modal component.
* Fixed by closing modal(s) on route change.


packages/components/modal/src/components/osds-modal/osds-modal.tsx
```diff
@@ -36,31 +34,35 @@ export class OsdsModal implements OdsModalAttribute, OdsModalMethod {
...
+  watchOpenedStateHandler(masked: boolean) {
     const directChildren = Array.from(document.body.children);
-    for (const child of directChildren) {
-      if (child !== this.el && child.nodeName !== 'SCRIPT') {
-        if (!this.masked) {
-          child.setAttribute('inert', '');
-        } else {
-          child.removeAttribute('inert');
-        }
+    const filteredChildren = directChildren.filter(child => child !== this.el && child.nodeName !== 'SCRIPT');
```
dpellier, line 51: not sure it's actually better as you go through 2 loops instead of one

packages/storybook/stories/components/modal/modal.web-components.stories.ts
```diff
@@ -69,9 +70,27 @@ const TemplateDefault = (args:any) => {
     }
   }
 
+  const locationChangeTrigger = () => {
```
dpellier, line 73: did you try instead to listen to the location change (with for eample a MutationObserver) instead of running a interval?
connect this watcher on page load

packages/components/modal/src/docs/osds-modal/usage.mdx
```diff
@@ -5,7 +5,22 @@ import GenericStyle from '@ovhcloud/ods-common-core/docs/generic-style.mdx';
 ## Usage
 
 ### Default
-<osds-modal>Modal</osds-modal>
+
+export const buttonTrigger = () => {
+  setTimeout(() => {
```
dpellier, line 10: why the setTimeout ?

packages/components/modal/src/components/osds-modal/osds-modal.tsx
manoncarbonnel: It felt like a separating concerns to me. But I can switch this part of the code back to its previous state

packages/components/modal/src/docs/osds-modal/usage.mdx
aesteves60: It's because the render of the page is not finish otherwise the selector is undefined

packages/storybook/stories/components/modal/modal.web-components.stories.ts
manoncarbonnel: That was a good advice ! I learned something today. Thanks. Code updated accordingly,

packages/components/modal/src/docs/osds-modal/usage.mdx
Leotheluck: can't we do it with an event listener of `DOMContentLoaded` instead ?

manoncarbonnel: I tried but it didn't lead to success.

packages/storybook/stories/components/modal/modal.web-components.stories.ts
```diff
@@ -59,6 +59,7 @@ const TemplateDefault = (args:any) => {
     const modal = document.querySelector('osds-modal');
     if (modal) {
       modal.open();
+      locationChangeTrigger();
```
dpellier, line 62: to remove

dpellier (changes requested):

dpellier (approved):

PR #224, opened 2023-09-28
fix(textarea): emit odsValueChange on value change

Leotheluck:
Fixing problem with Textarea component.

Event `odsValueChange` was not triggered and overall `value` was not updated on user input because of wrong data validation.

You can reproduce the issue by watching `odsValueChange` and typing inside the `textarea`. `odsValueChange` will only be triggered on `clear` or if `value` is equal to `''`.

The reason this problem bypassed the tests is `odsValueChange`'s emit was only tested on literal change of value, and not on reproducing the manual input behavior, hence the new test I added.

### DIFF
--- a/packages/components/textarea/src/components/osds-textarea/osds-textarea.tsx
+++ b/packages/components/textarea/src/components/osds-textarea/osds-textarea.tsx
@@ -130,39 +130,39 @@ export class OsdsTextArea implements OdsTextAreaAttribute, OdsTextAreaEvent, Ods
   /**
    * @see OdsTextAreaBehavior.beforeInit
    */
-   beforeInit(): void {
+  beforeInit(): void {
     this.controller.beforeInit();
   }
 
   /**
    * @see OdsTextAreaBehavior.emitChange
    */
-   emitChange(value: HTMLTextAreaElement['value'], oldValue?: HTMLTextAreaElement['value']): void {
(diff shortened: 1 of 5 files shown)


packages/components/textarea/src/components/osds-textarea/osds-textarea.tsx
```diff
@@ -130,39 +130,39 @@ export class OsdsTextArea implements OdsTextAreaAttribute, OdsTextAreaEvent, Ods
...
    * @see OdsTextAreaBehavior.emitChange
    */
-   emitChange(value: HTMLTextAreaElement['value'], oldValue?: HTMLTextAreaElement['value']): void {
+  emitChange(value: HTMLTextAreaElement['value'], oldValue?: HTMLTextAreaElement['value']): void {
     this.odsValueChange.emit({
-      value: value == null ? value : `${value}`,
       oldValue: oldValue == null ? oldValue : `${oldValue}`,
-      validity: this.controller.getTextAreaValidity()
+      validity: this.controller.getTextAreaValidity(),
+      value: value == null ? value : `${value}`,
```
aesteves60, line 144:
```suggestion
      value: value === null ? value : `${value}`,
```

aesteves60 (changes requested):

aesteves60 (approved):

astagnol (approved):

PR #227, opened 2023-10-03
fix(text): add missing test line for coverage

skhamvon:
Add missing tests for full coverage of text

### DIFF
--- a/packages/components/text/src/components/osds-text/osds-text.tsx
+++ b/packages/components/text/src/components/osds-text/osds-text.tsx
@@ -35,7 +35,7 @@ export class OsdsText implements OdsTextAttribute {
 
   render() {
     return (
-      <Host style={{ '--osds-text-color-specific-hue': `var(${odsGenerateColorVariable(this.color!, this.hue!)})` }}>
+      <Host style={{ '--osds-text-color-specific-hue': this.color && this.hue ? `var(${odsGenerateColorVariable(this.color, this.hue)})` : '' }}>
         <slot></slot>
       </Host>
     );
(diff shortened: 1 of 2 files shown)


packages/components/text/src/components/osds-text/osds-text.tsx
```diff
@@ -37,7 +37,7 @@ export class OsdsText implements OdsTextAttribute {
     return (
       <Host {...{
```
dpellier, line 38: would be nice to remove the unneccessary destructuring

skhamvon: It's removed

packages/components/text/src/components/osds-text/osds-text.tsx
```diff
@@ -35,11 +35,7 @@ export class OsdsText implements OdsTextAttribute {
 
   render() {
     return (
-      <Host {...{
-        style: {
-          '--osds-text-color-specific-hue': this.color && this.hue ? `var(${odsGenerateColorVariable(this.color, this.hue)})` : '',
-        }
-      }}>
+      <Host style={{ '--osds-text-color-specific-hue': `var(${odsGenerateColorVariable(this.color!, this.hue!)})` }}>
```
dpellier, line 38: We should avoid all `!` during code refacto.
Here the method should handle missing values. This method is only used by text so we can move it here instead of keeping it in theming.

Whole refacto would be:
- remove all trace of `odsGenerateColorVariable` from `ods-common-theming`
- add a local function on `osds-text` component that does the same except that it return empty string if one or both arguments are missing
- add unit test to validate the function return either empty of correct string with different args
- remove the `!`

dpellier (changes requested):

packages/components/text/src/components/osds-text/osds-text.tsx
skhamvon: Ternary is restored and a new test has been added
odsGenerateColorVariable may be used on another component, so it was recommend to let it in common-theming for now

dpellier (approved):

PR #232, opened 2023-10-05
fix(datepicker): fix exports

Leotheluck:
> Adding exports for `odsDatepickerValueChange`, `odsDatepickerValueChangeEventDetail`, `ODS_DATEPICKER_DAY` & `ODS_DATEPICKER_DAYS` inside `public-api.ts`

> Adding `build:ci` command

> Adding `formattedValue` inside `odsDatepickerValueChange` event.

### DIFF
--- a/packages/components/datepicker/src/components/osds-datepicker/osds-datepicker.tsx
+++ b/packages/components/datepicker/src/components/osds-datepicker/osds-datepicker.tsx
@@ -265,6 +265,7 @@ export class OsdsDatepicker implements OdsDatepickerAttribute, OdsDatepickerEven
           placeholder={placeholder}
           type={ODS_INPUT_TYPE.text}
           value={this.formatDate(value)}
+          style={{ '--placeholder-color': 'red' }}
         ></osds-input>
         <input
           tabindex={-1}
--- a/packages/components/datepicker/src/components/osds-datepicker/osds-datepicker.tsx
+++ b/packages/components/datepicker/src/components/osds-datepicker/osds-datepicker.tsx
@@ -265,7 +265,6 @@ export class OsdsDatepicker implements OdsDatepickerAttribute, OdsDatepickerEven
           placeholder={placeholder}
           type={ODS_INPUT_TYPE.text}
           value={this.formatDate(value)}
-          style={{ '--placeholder-color': 'red' }}
         ></osds-input>
         <input
           tabindex={-1}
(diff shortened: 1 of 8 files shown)


packages/components/datepicker/src/components/osds-datepicker/osds-datepicker.tsx
```diff
@@ -119,7 +119,7 @@ export class OsdsDatepicker implements OdsDatepickerAttribute, OdsDatepickerEven
   }
 
   emitDatepickerValueChange(newValue: Date | undefined | null, oldValue?: Date | undefined | null): void {
-    this.odsDatepickerValueChange.emit({ value: newValue, oldValue: oldValue });
+    this.odsDatepickerValueChange.emit({ value: newValue, oldValue: oldValue, formattedValue: newValue && this.format ? Datepicker.formatDate(newValue, this.format) : undefined });
```
dpellier, line 122: please some tests about this

dpellier (changes requested):

packages/components/datepicker/src/components/osds-datepicker/osds-datepicker.tsx
```diff
@@ -265,6 +265,7 @@ export class OsdsDatepicker implements OdsDatepickerAttribute, OdsDatepickerEven
           placeholder={placeholder}
           type={ODS_INPUT_TYPE.text}
           value={this.formatDate(value)}
+          style={{ '--placeholder-color': 'red' }}
```
dpellier, line 268: can't you use a class instead to keep css code outside of ts code?
replace 'red' with an actual value from the theme instead

dpellier (changes requested):

packages/components/datepicker/src/components/osds-datepicker/osds-datepicker.tsx
```diff
@@ -24,9 +24,14 @@ export class OsdsDatepicker implements OdsDatepickerAttribute, OdsDatepickerEven
   @Element() el!: HTMLElement;
 
   @State() hasFocus = false;
-  @State() datepickerInstance: Datepicker | undefined = undefined;
-  @State() hiddenInput: HTMLInputElement | undefined = undefined;
-  @State() datepickerElement: HTMLElement | undefined = undefined;
+
+  protected datepickerInstance: Datepicker | undefined = undefined;
```
dpellier, line 28: you can use `private` directly as the controller is not a subclass so he can't access it

dpellier (changes requested):
I see you're using `flatMap` in the code.
You need to update the root `tsconfig.json` `lib` like:
```
"lib": ["dom", "es2018", "esnext.array", "es2019"],
```

dpellier (approved):

PR #235, opened 2023-10-09
feat(text): use local import instead of theme directly

skhamvon:
Use local component import instead of importing from common-theming to uniform way of import


aesteves60 (changes requested):
You forgot to add the new enum in the public-api.ts

dpellier (approved):

aesteves60 (approved):

PR #236, opened 2023-10-09
fix(modal): trigger inert at modal open and fix e2e test

manoncarbonnel:
Fixed end-to-end tests for modal component:

- trigger inert at modal open (masked is false on component create and masked is also false on open, so it did not trigger the inert watcher)
- fix e2e test by adding open to the test for inert attribute

### DIFF
--- a/packages/components/modal/src/components/osds-modal/osds-modal.tsx
+++ b/packages/components/modal/src/components/osds-modal/osds-modal.tsx
@@ -43,7 +43,6 @@ export class OsdsModal implements OdsModalAttribute, OdsModalMethod {
   @Method()
   async open(): Promise<void> {
     this.masked = false;
-    this.inertBodyChildren(this.masked);
   }
 
   @Watch('masked')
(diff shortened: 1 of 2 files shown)


packages/components/modal/src/components/osds-modal/osds-modal.tsx
```diff
@@ -43,16 +43,21 @@ export class OsdsModal implements OdsModalAttribute, OdsModalMethod {
   @Method()
   async open(): Promise<void> {
     this.masked = false;
+    this.inertBodyChildren(this.masked);
```
dpellier, line 46: changing the `masked` Prop value will trigger the `masked` Watch, so the inert change should only be executed once (in the Watch handler)

dpellier (changes requested):

packages/components/modal/src/components/osds-modal/osds-modal.tsx
manoncarbonnel: Problem was ["The `@Watch()` decorator does not fire when a component initially loads."](https://stenciljs.com/docs/reactive-data#the-watch-decorator-watch).

dpellier (approved):

PR #241, opened 2023-10-16
Feat/datagrid step 4

aesteves60:
Add custom formatter by column

### DIFF
--- a/packages/components/datagrid/src/components/osds-datagrid/core/controller.ts
+++ b/packages/components/datagrid/src/components/osds-datagrid/core/controller.ts
@@ -11,6 +11,12 @@ class OdsDatagridController {
         private readonly component: OsdsDatagrid,
     ) {}
 
+    validateAttributes(): void {
+        if (!this.component.height) {
+            this.component.logger.warn('Attributes "height" need to be set');
+        }
+    }
+
     toTabulatorColumn(column: OdsDatagridColumn): ColumnDefinition {
         return {
             title: column.title,
--- a/packages/components/datagrid/src/components/osds-datagrid/core/controller.ts
+++ b/packages/components/datagrid/src/components/osds-datagrid/core/controller.ts
@@ -17,8 +17,9 @@ class OdsDatagridController {
             field: column.field,
             headerSort: column.isSortable ?? false,
             headerHozAlign: 'center',
+            vertAlign: 'middle',
             titleFormatter: (cell: CellComponent) => this.getOdsText(cell.getValue(), ODS_TEXT_SIZE._500),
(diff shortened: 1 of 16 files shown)


packages/components/datagrid/src/components/osds-datagrid/core/controller.ts
```diff
@@ -17,8 +17,9 @@ class OdsDatagridController {
             field: column.field,
             headerSort: column.isSortable ?? false,
             headerHozAlign: 'center',
+            vertAlign: 'middle',
             titleFormatter: (cell: CellComponent) => this.getOdsText(cell.getValue(), ODS_TEXT_SIZE._500),
-            formatter: (cell: CellComponent) => this.getOdsText(cell.getValue(), ODS_TEXT_SIZE._400),
+            formatter: (cell: CellComponent) => column.formatter?.(cell.getValue()) ?? this.getOdsText(cell.getValue(), ODS_TEXT_SIZE._400),
```
dpellier, line 22: would be nice to add a dedicated sandbox page with a very large datagrid (10000+ rows) to ensure we don't have any perf issue

dpellier (changes requested):

dpellier (changes requested):
Also add a `build:ci` command on datagrid package otherwise doc will not be build

dpellier (changes requested):
Cursor is always `pointer` on each row, should be default one until row get interactive (like draggable)

feoche (changes requested):
UI review : 
- [x] In "Default" Storybook story, `rowHeight` property should be 52px (3.25rem)
- [x] Changing story parameters does not trigger component re-render
- [x] Cursor should be `auto` when hovering rows
- [ ] [Safari] column with icon is 1px off at the bottom of each cell (can be fixed with --webkit-inline-flex on row)
- [x] Sortable column label height is different to non-sortable label
- [x] Header row has different height than the `rowHeight` (to be discussed)

- [x] (Switching stories between "Default" and "Large Data" breaks its display - hard to reproduce)
- [x] Remove control on Large Data Page

Leotheluck (approved):

astagnol (changes requested):
Click on isSelectable control, then click again. Header is broken:
Capture d’écran 2023-10-24 à 17 40 54

feoche:
- [ ] Removing all columns does not get the "noResultLabel" to appear
- [ ] Header is always as "fixed", without a drop shadow (to discuss with Design)

astagnol (approved):

feoche (approved):

PR #250, opened 2023-10-23
refactor(icon): style to BEM & 1 scss file

aesteves60:


packages/components/icon/src/components/osds-icon/osds-icon.tsx
```diff
@@ -56,10 +53,14 @@ export class OsdsIcon implements OdsIconAttribute {
     const base64Icon = icons[this.name];
     return (
       <Host {...{
```
dpellier, line 55: can you also remove the destructuring when not necessary ?

dpellier (changes requested):

dpellier (approved):

Leotheluck (approved):

PR #251, opened 2023-10-24
refactor(flag): clean up && put screenshot on one page

dpellier:
- style using BEM
- e2e screenshot on one page
Assets loading can not be change for now, so minimal clean-up

### DIFF
--- a/packages/components/flag/src/components/osds-flag/core/controller.ts
+++ b/packages/components/flag/src/components/osds-flag/core/controller.ts
@@ -6,14 +6,10 @@ import {
   OdsLoadContent,
   OdsSvgValidator,
   odsGetSrc,
-  odsGetAssetPath
+  odsGetAssetPath,
 } from '@ovhcloud/ods-common-core';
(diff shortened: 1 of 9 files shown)


packages/components/flag/src/components/osds-flag/core/controller.ts
```diff
@@ -31,7 +27,7 @@ class OdsFlagController {
    * @param isVisible - is the element to load is visible or not
    * @param isBrowser - is the context is browser one or not
    */
-  load(isVisible: boolean, isBrowser = true): Promise<string | void> {
+  async load(isVisible: boolean, isBrowser = true): Promise<string | void> {
```
aesteves60, line 30: No await used

aesteves60 (changes requested):

packages/components/flag/src/components/osds-flag/core/controller.ts
dpellier: It's just a best practice to show when a function returns a Promise

aesteves60 (approved):

PR #255, opened 2023-10-26
fix(datepicker): close picker when select date

aesteves60:


packages/components/datepicker/src/components/osds-datepicker/core/controller.ts
```diff
@@ -29,6 +29,8 @@ class OdsDatepickerController {
         this.component.datepickerInstanceAccessor?.setDate(newValue);
         this.component.emitDatepickerValueChange(newValue, oldValue ? oldValue : null);
       }
+      this.component.el.focus();
```
dpellier, line 32: can you add a comment with the explanation why both call are necessary

dpellier (changes requested):

packages/components/datepicker/src/components/osds-datepicker/core/controller.ts
aesteves60: I don't know why

manoncarbonnel (approved):

dpellier (approved):

PR #259, opened 2023-10-30
fix(button): not ghost variant on circle button

aesteves60:
Remove ghost styling on circle button


dpellier (changes requested):
There are no hover and focus anymore on circle button, regarding the variant, is that really on purpose.
No sure about what was asked here as current prod circle button seemes to be the same on all variant (ghost, flat, stroked)

dpellier (approved):

PR #260, opened 2023-10-30
fix(tile): fixing display on Safari browser by adding flex properties

manoncarbonnel:
Fixing display on Safari browser by adding flex properties

### DIFF
--- a/packages/components/tile/src/components/osds-tile/osds-tile.scss
+++ b/packages/components/tile/src/components/osds-tile/osds-tile.scss
@@ -7,7 +7,9 @@
 
 :host {
   box-sizing: border-box;
+  display: flex;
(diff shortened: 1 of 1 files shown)


packages/components/tile/src/components/osds-tile/osds-tile.scss
```diff
@@ -7,7 +7,9 @@
 
 :host {
   box-sizing: border-box;
+  display: flex;
   flex-direction: row;
+  justify-content: space-between;
```
dpellier, line 12: hum not sure we should enforce this render by defaut, is that on the compo spec requirement?

dpellier (approved):

aesteves60 (approved):

PR #266, opened 2023-11-03
fix(accordion): dont open when disabled

skhamvon:
- Fix accordion that was opening even when it was disabled

### DIFF
--- a/packages/components/accordion/src/components/osds-accordion/core/controller.ts
+++ b/packages/components/accordion/src/components/osds-accordion/core/controller.ts
@@ -27,8 +27,10 @@ class OdsAccordionController {
    * in case of no `details` already set, it won't change the `opened` status.
    */
   onToggle(): void {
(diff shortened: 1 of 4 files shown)


aesteves60 (approved):

aesteves60 (changes requested):
Need test

aesteves60 (approved):

astagnol (changes requested):
Working but pending since we need a final check about this fix with design team.
And I think it could be nice to add 'disabled' in Storybook controls

dpellier (changes requested):
can you add a `disable` control on stroybook for this component?

dpellier (changes requested):
Disabled accordion is still focusable, is it the expected behaviour?

dpellier (approved):

aesteves60 (approved):

astagnol (approved):

PR #268, opened 2023-11-03
fix(datepicker): hideSibligsMonthDays on update datepicker

aesteves60:


packages/components/datepicker/src/components/osds-datepicker/osds-datepicker.tsx
```diff
@@ -147,6 +147,7 @@ export class OsdsDatepicker implements OdsDatepickerAttribute, OdsDatepickerEven
         ? this.minDate
         : undefined,
     });
+    this.hideSiblingsMonthDays();
```
dpellier, line 150: maybe this should be renamed as it does not necessarily hide things, something like `updateSiblingsMonthDaysVisibility`?

dpellier (changes requested):

dpellier (approved):

Leotheluck (approved):

PR #269, opened 2023-11-06
fix(quantity): change value onblur to be valid with step

skhamvon:
ODS-784
Change value of quantity component to match steps

### DIFF
--- a/packages/components/quantity/src/components/osds-quantity/core/controller.spec.ts
+++ b/packages/components/quantity/src/components/osds-quantity/core/controller.spec.ts
@@ -123,6 +123,24 @@ describe('spec:ods-quantity-controller', () => {
           expect(input.value).toEqual(input.max);
         });
 
+        it('should change value of input if not in a valid step', () => {
+          setup();
+          let validInput = "1";
+          const input = document.createElement('input');
+          input.setAttribute('type', 'number');
+          input.setAttribute('value', '1.5');
+          input.setAttribute('min', '0');
+          input.setAttribute('step', '1');
+
+          component.el.appendChild(input);
+
+          controller.initInput();
+          input.focus();
+          input.blur();
+
+          expect(input.value).toEqual(validInput);
+        });
+
         it('should change value of osds-input if inferior to min on Blur', () => {
           setup();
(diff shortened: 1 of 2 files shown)


aesteves60 (approved):

packages/components/quantity/src/components/osds-quantity/core/controller.spec.ts
```diff
@@ -123,6 +123,24 @@ describe('spec:ods-quantity-controller', () => {
...
+          input.setAttribute('min', '0');
+          input.setAttribute('step', '1');
+
+          component.el.appendChild(input);
+
+          controller.initInput();
+          input.focus();
+          input.blur();
+
+          expect(input.value).toEqual(validInput);
```
dpellier, line 141: `toBe` when comparing primitive value

astagnol (approved):

packages/components/quantity/src/components/osds-quantity/core/controller.ts
```diff
@@ -35,11 +35,19 @@ class OdsQuantityController {
...
+      if (stepNb && !Number.isInteger((valueAsNb - minNb) / stepNb)) {
+        valueToCheck = Math.floor( valueAsNb / stepNb ) * stepNb
+      }
+
+      if (this.component.input.min !== '' && valueToCheck < minNb) {
         this.component.input.value = minNb;
-      } else if (this.component.input.max !== '' && valueAsNb > maxNb) {
+      } else if (this.component.input.max !== '' && valueToCheck > maxNb) {
         this.component.input.value = maxNb;
+      } else if (valueToCheck != valueAsNb) {
```
dpellier, line 49: should be `!==`
if this doesn't work then there is a type issue, but we shouldn' compare string and number value and expect it to work correctly

dpellier (changes requested):

packages/components/quantity/src/components/osds-quantity/core/controller.ts
skhamvon: Fixed

dpellier (approved):

PR #272, opened 2023-11-08
feat(tile): use BEM style

skhamvon:


Leotheluck (approved):

packages/components/tile/src/components/osds-tile/osds-tile.tsx
```diff
@@ -65,13 +65,23 @@ export class OsdsTile implements OdsTileAttribute {
     this.beforeRender();
   }
 
+
```
aesteves60, line 68: to remove

packages/components/tile/src/components/osds-tile/osds-tile.scss
```diff
@@ -5,112 +5,114 @@
...
+      grid-template-columns: auto 1fr auto;
+      width: 100%;
+    }
+  }
 
-      100% {
-        background-position: 0 0;
-      }
+  :host(&#{&}--rounded){
+    @include osds-tile-on-selected-host {
```
aesteves60, line 105: Not sure but you have a double host(&) here

aesteves60 (changes requested):

packages/components/tile/src/components/osds-tile/osds-tile.tsx
skhamvon: removed

packages/components/tile/src/components/osds-tile/osds-tile.scss
skhamvon: Isn't this suppose to mean :
:host(.osds-tile.osds-tile--rounded) ?

dpellier: you don't need to duplicate the selector
`:host(.osds-tile--rounded)` is enough

skhamvon: It's changed

dpellier (approved):

aesteves60 (approved):

PR #274, opened 2023-11-08
feat(medium): init component

manoncarbonnel:
Adding a new component called Medium


packages/components/medium/src/components/osds-medium/core/controller.spec.ts
```diff
@@ -0,0 +1,127 @@
...
+
+class OdsMediumMock extends OsdsMedium {
+  constructor(attribute: Partial<OsdsMedium>) {
+    super();
+    Object.assign(this, attribute);
+  }
+
+  controller: OdsMediumController = jest.fn() as unknown as OdsMediumController;
+  odsRemoveClick!: EventEmitter<void>;
+  removeClicked = jest.fn();
```
dpellier, line 17: remove

dpellier (changes requested):
no need to put number on file names in `packages/storybook/stories/components/medium`

dpellier (changes requested):
Please update storybook controls so that we can update width and height
Capture d’écran 2023-11-08 à 17 13 48

packages/components/medium/src/components/osds-medium/core/controller.ts
```diff
@@ -0,0 +1,67 @@
...
+  constructor(component: OsdsMedium) {
+    this.component = component;
+  }
+
+  /**
+   * Warn user if the alt attribute is not set
+   * @param alt - alternative description of the medium
+   */
+  validateAlt(alt: string): void {
+    if (alt.length == 0) {
```
aesteves60, line 21:
```suggestion
    if (alt.length === 0) {
```

packages/components/medium/src/components/osds-medium/core/controller.ts
```diff
@@ -0,0 +1,67 @@
...
+      this.logger.warn('The alt attribute must be set');
+    }
+  }
+
+  /**
+   * Warn user if the src attribute is not set
+   * @param src - url of the medium
+   */
+  validateSrc(src: string): void {
+    if (src.length == 0) {
```
aesteves60, line 31:
```suggestion
    if (src.length === 0) {
```

packages/components/medium/src/components/osds-medium/osds-medium.tsx
```diff
@@ -0,0 +1,63 @@
...
+ */
+@Component({
+  shadow: true,
+  styleUrl: 'osds-medium.scss',
+  tag: 'osds-medium',
+})
+export class OsdsMedium implements OdsMediumAttribute {
+  controller: OdsMediumController = new OdsMediumController(this);
+
+  /** @see OdsMediumAttributes.alt */
```
aesteves60, line 18: I think this type of comment are useless, not utils in the doc generated

packages/components/medium/src/components/osds-medium/osds-medium.tsx
```diff
@@ -0,0 +1,63 @@
...
+  connectedCallback(): void {
+    this.beforeInit();
+  }
+
+  render(): FunctionalComponent {
+    return (
+      <Host>
+        <img alt={ this.alt }
+          src={ this.src }
+          {...(this.height ? { height: this.height } : {}) }
```
aesteves60, line 57: Don't work with a undefined fallback ? For avoid the double deconstruction
like that `height={this.height || undefined}`

aesteves60 (changes requested):

packages/components/medium/src/components/osds-medium/osds-medium.tsx
dpellier: Technically `0` should be a valid value, in that case it won't be

dpellier: use `??` instead

dpellier (approved):

Leotheluck (approved):

aesteves60 (approved):

PR #279, opened 2023-11-14
fix(phone-number): trigger event

aesteves60:
Trigger event odsValueChange everytime when type on the input


dpellier (changes requested):
also need to manage the case when the component is initialized with a value, we need to trigger a update event with the correct validity (but only if the value is not empty)

astagnol (approved):

Leotheluck (approved):

PR #281, opened 2023-11-15
fix(phone-number): indicator included validation

skhamvon:
ODS-1007
Change input to error when indicator is repeated within the value


dpellier (changes requested):
Hum, this issue should have been closed and should not be considered

PR #285, opened 2023-11-16
fix(modal): create wrapper to handle modal body append

skhamvon:
ODS-1001
- Create a wrapper and locate the modal within
- When opening the modal, it is relocated to body to allow the inert to work correctly
- Relocate the modal to its wrapper once closed

- Fix contrasted issue on the close icon


Leotheluck (changes requested):
Doesn't work, makes the modal inert. See storybook

feoche:
> Doesn't work, makes the modal inert. See storybook

Opening the modal in Storybook at start does not move the DOM component outside of its wrapper

skhamvon:
> > Doesn't work, makes the modal inert. See storybook
> 
> Opening the modal in Storybook at start does not move the DOM component outside of its wrapper

Fixed

PR #287, opened 2023-11-20
fix(phone-number): flag height

aesteves60:
Remove console.log
Set flag height & width the same


Leotheluck (changes requested):
The problem is now on the vertical aspect, far less noticeable though
Capture d’écran 2023-11-20 à 14 19 30

PR #288, opened 2023-11-21
fix(datagrid): row change & storybook

aesteves60:
Error on row change: test throx when the fix was not present
Phone-number / Search-bar select cropped


astagnol (changes requested):
Datagrid is not displayed as expected in usage guidelines when using a Menu with row formatter:
Capture d’écran 2023-11-22 à 10 08 50

It seems there is still a regression from v16.2.1 (Select):
Capture d’écran 2023-11-22 à 10 21 05
Not sure this regression is related to Datagrid though. We can discuss about it

feoche (approved):

astagnol (approved):

Leotheluck (approved):

PR #293, opened 2023-11-23
[v18] Form v18

skhamvon:
ODS-1088
Move common attributes of multiples form input component to common core


feoche (changes requested):
Rename all "form-field" changes into "field"

packages/common/core/src/form/constants/ods-generic-form-field-size.ts
astagnol, line 1: Since Form field component exists, maybe we should rename this to "field" (or something else) instead of "form-field"?

astagnol (changes requested):

skhamvon:
> Rename all "form-field" changes into "field"

Changed

packages/common/core/src/form/interfaces/attributes.ts
```diff
@@ -0,0 +1,71 @@
+import type { ODS_COMMON_FIELD_SIZE } from '../constants/ods-common-field-size';
+import type { ODS_INPUT_TYPE } from '../constants/ods-input-type';
```
dpellier, line 2: should be renamed `ODS_COMMON_INPUT_TYPE` also

packages/components/input/src/components/osds-input/core/controller.ts
```diff
@@ -70,7 +67,7 @@ class OdsInputController {
   private hasForbiddenValue(): boolean {
     switch (this.component.type) {
     case 'number':
-      return this.component.forbiddenValues.some((forbiddenValue) => {
+      return this.component.forbiddenValues.some((forbiddenValue: any) => {
```
dpellier, line 70: prefer `unknown` to `any`

packages/components/input/src/components/osds-input/core/controller.ts
```diff
@@ -80,7 +77,7 @@ class OdsInputController {
         return false;
       });
     default:
-      return this.component.forbiddenValues.some((forbiddenValue) => {
+      return this.component.forbiddenValues.some((forbiddenValue: any) => {
```
dpellier, line 80: prefer `unknown` to `any`

packages/components/input/src/components/osds-input/constants/input-size.ts
```diff
@@ -1,10 +0,0 @@
-enum ODS_INPUT_SIZE {
```
dpellier, line 1: keep exporting `ODS_INPUT_SIZE` as a wrapper of `ODS_COMMON_FIELD_SIZE`
same for `ODS_INPUT_TYPE`
(see text component constants for example)

the goal is to provide those to the end user, but not to provide any direct reference to ods-common-core

packages/common/core/src/form/interfaces/attributes.ts
skhamvon: Done

packages/components/input/src/components/osds-input/core/controller.ts
skhamvon: Done

packages/components/input/src/components/osds-input/core/controller.ts
skhamvon: Done

packages/components/input/src/components/osds-input/constants/input-size.ts
skhamvon: Done

packages/storybook/stories/components/input/3_demo.stories.ts
```diff
@@ -1,5 +1,5 @@
 import { ODS_THEME_COLOR_INTENTS } from '@ovhcloud/ods-common-theming';
-import { ODS_INPUT_SIZES, ODS_INPUT_TYPES } from '@ovhcloud/ods-components/input';
+import { ODS_COMMON_FIELD_SIZES, ODS_COMMON_INPUT_TYPES } from '@ovhcloud/ods-common-core';
```
dpellier, line 2: should be `import { ODS_INPUT_SIZES, ODS_INPUT_TYPES } from '@ovhcloud/ods-components/input';` here ;)
the story behave the same as a user who would use the input component (so it won't know about common-core)

dpellier (changes requested):

manoncarbonnel (approved):

Leotheluck (approved):

packages/storybook/stories/components/input/3_demo.stories.ts
```diff
@@ -17,7 +16,7 @@ const storyParams = {
   },
   type: {
     category: 'General',
-    defaultValue: ODS_INPUT_TYPE.text,
+    defaultValue: 'text',
```
astagnol, line 19: Capture d’écran 2023-12-22 à 10 10 14
We should be able to select a type

packages/storybook/stories/components/input/3_demo.stories.ts
```diff
@@ -27,16 +26,8 @@ const storyParams = {
...
-  },
-  value: {
-    category: 'General',
-    control: { type: 'text' },
-  },
-  icon: {
-    category: 'General',
-    defaultValue: '',
-    options: ODS_ICON_NAMES,
+    defaultValue: 'md',
```
astagnol, line 29: Capture d’écran 2023-12-22 à 10 23 16
Something went wrong in this file. 
Size is no longer displayed as control in Input story so, it should be removed here.
And we should still be able to play with value and icon in the controls.

packages/storybook/stories/components/input/3_demo.stories.ts
```diff
@@ -139,3 +130,65 @@ type ValidationProps = {
   ...(extractStoryParams(storyParams) as Record<string, unknown>),
 };
 (Validation as unknown as ValidationProps).play = InputPlay;
+
+/* All Inputs */
```
astagnol, line 134: This part has been removed in Storybook refactor step 3. 
Please remove "All Inputs" example here since we defined that only components like Flag and Icon should keep an "All" story (usage is described in Usage Guidelines)

astagnol (changes requested):

packages/common/core/src/form/method.controller.ts
```diff
@@ -0,0 +1,125 @@
...
+  constructor(private readonly component: OdsCommonFieldComponent) { }
+
+  /**
+   * empty the value
+   */
+  async clear(): Promise<void> {
+    if (this.component.disabled) {
+      return;
+    }
+    this.component.value = '';
```
dpellier, line 32: should we reset to `undefined` or `null` instead to avoid having string value on wrong input (like `type="number"`)

packages/common/core/src/form/method.controller.ts
```diff
@@ -0,0 +1,125 @@
...
+  /**
+   * active the focus on the input in order to let the user write something
+   */
+  async setFocus(): Promise<void> {
+    if (this.component.disabled) {
+      return;
+    }
+    this.component.inputEl?.focus();
+    this.component.hasFocus = true;
+    this.component?.emitFocus?.();
```
dpellier, line 47: no `?` needed on component

packages/common/core/src/form/method.controller.ts
```diff
@@ -0,0 +1,125 @@
...
+
+  /**
+   * active the focus on the input in order to let the user write something
+   */
+  async setFocus(): Promise<void> {
+    if (this.component.disabled) {
+      return;
+    }
+    this.component.inputEl?.focus();
+    this.component.hasFocus = true;
```
dpellier, line 46: seems weird to set the `hasFocus` to true if the previous line fails (ie, no `inputEl` found)

packages/common/core/src/form/method.controller.ts
```diff
@@ -0,0 +1,125 @@
...
+      return;
+    }
+    this.component.masked = !this.component.masked;
+  }
+
+  /**
+   * restore the value to the initial state
+   */
+  async reset(): Promise<void> {
+    this.component.value = this.component.defaultValue ? this.component.defaultValue.toString() : '';
```
dpellier, line 64: same concern as above above wrong type as we enforce `string`

packages/common/core/src/form/interfaces/attributes.ts
```diff
@@ -0,0 +1,73 @@
...
+  /** Placeholder text for the input */
+  placeholder?: string;
+  /** Text before the input value */
+  prefixValue?: string;
+  /** Indicates if the input is read-only or not */
+  readOnly?: boolean;
+  /** Indicates if the input is required or not */
+  required?: boolean;
+  /** Size of the input: see component principles */
+  size: ODS_COMMON_FIELD_SIZE;
```
dpellier, line 59: why required?

dpellier (changes requested):

packages/common/core/src/form/method.controller.ts
```diff
@@ -55,13 +49,15 @@ class OdsCommonFieldMethodController implements OdsCommonFieldMethod {
...
     this.component.masked = !this.component.masked;
+    this.component.odsHide.emit();
   }
 
   /**
    * restore the value to the initial state
    */
   async reset(): Promise<void> {
-    this.component.value = this.component.defaultValue ? this.component.defaultValue.toString() : '';
+    this.component.value = this.component.defaultValue ? this.component.defaultValue.toString() : null;
```
dpellier, line 59: `toString` needed? will "break" number value type

dpellier (changes requested):

packages/common/core/src/form/method.controller.ts
```diff
@@ -34,51 +48,19 @@ class OdsCommonFieldMethodController<T extends OdsCommonFieldAttribute & OdsComm
...
-        return false;
-      }) || false;
-    default:
-      return this.component.forbiddenValues?.some((forbiddenValue: OdsInputValue) => {
-        return forbiddenValue === this.component.value;
-      }) || false;
-    }
+  async reset(): Promise<void> {
+    this.component.value = this.component.defaultValue ?? null;
+    this.component.odsReset?.emit();
```
dpellier, line 60: `?` probably not needed anymore

packages/components/src/input/src/components/osds-input/osds-input.tsx
```diff
@@ -36,49 +37,45 @@ export class OsdsInput implements OdsInputAttribute, OdsInputEvent, OdsInputMeth
...
+  @Prop({ reflect: true }) readOnly?: boolean;
 
-  @Prop({ reflect: true }) size?: ODS_COMMON_FIELD_SIZE = DEFAULT_ATTRIBUTE.size;
+  @Prop({ reflect: true }) required?: boolean;
 
-  @Prop({ reflect: true }) step?: number = DEFAULT_ATTRIBUTE.step;
+  @Prop({ reflect: true }) step?: number;
 
-  @Prop({ reflect: true }) type?: ODS_COMMON_INPUT_TYPE = DEFAULT_ATTRIBUTE.type;
+  @Prop({ reflect: true }) type?: ODS_INPUT_TYPE = DEFAULT_ATTRIBUTE.type;
```
dpellier, line 78: BTW you can remove type when using DEFAULT_ATTR, Stencil v4.x+ now infer correctly

packages/common/core/src/form/method.controller.ts
```diff
@@ -0,0 +1,66 @@
...
+    }
+    this.component.value = null;
+    this.component.odsClear?.emit();
+  }
+
+  async getValidity<T extends { validity: ValidityState }>(element: T): Promise<OdsCommonFieldValidityState> {
+    const forbiddenValue = this.hasForbiddenValue();
+    const isValid = forbiddenValue ? false : element.validity.valid;
+    return {
+      ...element.validity,
```
dpellier, line 25: this won't work as ValidityState values seems to not be in the ob ject but in its `prototype` (so things like `Object.keys` for example won't work also).
You have to manually loop using `for .. in` to read the actual value adn generate a copy object

dpellier (changes requested):

packages/common/core/src/form/constants/ods-common-field-size.ts
```diff
@@ -0,0 +1,10 @@
+enum ODS_COMMON_FIELD_SIZE {
```
dpellier, line 1: remove size completely

dpellier (changes requested):

packages/components/src/input/src/components/osds-input/osds-input.tsx
aesteves60: I have an error on the component.d.ts without the type. The import is missing the the components.d.ts

packages/common/core/src/form/method.controller.ts
aesteves60: Not like the `for..in` because of the ts-ignore & the cast of the type (`as unknown as ValidityState`). So I create a mapper

packages/common/core/src/form/method.controller.ts
```diff
@@ -2,10 +2,23 @@ import type { OdsCommonFieldAttribute, OdsCommonFieldValidityState } from './int
...
 class OdsCommonFieldMethodController<T extends OdsCommonFieldAttribute & OdsCommonFieldEvent> implements OdsCommonFieldMethod {
+  protected logger: OdsLogger;
 
-  constructor(protected readonly component: T) { }
+  constructor(protected readonly component: T, loggerContext: string) {
+    this.logger = new OdsLogger(loggerContext);
+   }
+
+  beforeInit() {
+    OdsWarnComponentAttribute<string, T>({
```
dpellier, line 16: I would prefer that we don't rely on ODS logger at all anymore

dpellier (changes requested):

dpellier (changes requested):
Input clearable is still focusable when disabled
- go to demo
- set clearable and disabled to true
- tab => focus is on the clear button
(the issue is not on master branch)

PR #300, opened 2023-11-27
refactor(formfields): use attributes from common core

skhamvon:
ODS-1089
Use common core attributes on these component :
- checkbox
- datepicker
- password


packages/components/checkbox/src/components/osds-checkbox/interfaces/attributes.ts
```diff
@@ -1,6 +1,8 @@
+import type { OdsGenericFormFieldAttribute } from '@ovhcloud/ods-common-core';
+
 type OdsCheckboxAttributeCbk = ({ checked, value }: { checked: boolean, value: HTMLInputElement['value'] }) => Promise<void>;
 
-interface OdsCheckboxAttribute {
+interface OdsCheckboxAttribute extends OdsGenericFormFieldAttribute {
```
astagnol, line 5: Same here, Form field component exists, so maybe we should not use the same naming

astagnol (changes requested):

packages/components/checkbox/src/components/osds-checkbox/interfaces/attributes.ts
skhamvon: Name changed

PR #301, opened 2023-11-27
feat(chip): add odsChipRemoval event

Leotheluck:
- Added `odsChipRemoval` event
- If `OsdsChip` is removable and is clicked, it will emit event `odsChipRemoval`
- If `OsdsChip` is removable, focused on and key `Enter` is pressed, it will emit event `odsChipRemoval`


astagnol (changes requested):
Rebase needed

astagnol (changes requested):
Fix unit tests

dpellier (changes requested):
Chip behavior feels a bit weird regarding outline.
If you focus (using tab), you see the outline, then it's ok to emit a close event on Enter.
But if you just click on the chip, there are no outline, but still the close event is emitted on Enter.
@astagnol ?

astagnol (approved):

PR #303, opened 2023-12-04
fix(modal): change overflow to allow ocdk-surface correct display

skhamvon:
ODS-1119
Change overflow of the content class to allow a better display when a surface is present in the modal

### DIFF
--- a/packages/components/src/modal/src/components/osds-modal/osds-modal.scss
+++ b/packages/components/src/modal/src/components/osds-modal/osds-modal.scss
@@ -50,10 +50,12 @@
     }
 
     .content {
+      max-height: 30vh;
+      overflow-y: auto;
+
       .slot {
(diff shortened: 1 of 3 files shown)


packages/components/modal/src/components/osds-modal/osds-modal.scss
```diff
@@ -50,10 +50,12 @@
     }
 
     .content {
+      max-height: 30vh;
```
dpellier, line 53: is this a design prerequisite?

skhamvon: Yes !
You can see it in comment in the task

dpellier (approved):

aesteves60 (approved):

Leotheluck (approved):

PR #314, opened 2023-12-12
Refactor/input evnts

skhamvon:
Move input events interfaces to ods-common-core


packages/common/core/src/form/interfaces/events.ts
```diff
@@ -0,0 +1,25 @@
+import type { OdsValidityState } from '../validation/ods-validity-state';
+import type { EventEmitter } from '@stencil/core';
+
+interface OdsCommonFieldValueChangeEventDetail {
```
aesteves60, line 4: We need the name of the field too

packages/components/src/input/src/components/osds-input/interfaces/events.ts
```diff
@@ -1,22 +1,14 @@
...
-  value: string | undefined | null;
-  oldValue?: string | undefined | null;
-}
-
-type OdsInputValueChangeEvent = CustomEvent<OdsInputValueChangeEventDetail>;
-
-interface OdsInputEvent {
-  odsInputBlur: EventEmitter<void>;
-  odsInputFocus: EventEmitter<void>;
+interface OdsInputEvent extends OdsCommonFieldEvent {
```
aesteves60, line 4: Don't rewrite the `odsBlur`, `odsFocus` & `odsValueChange`

aesteves60 (changes requested):
Need to add the event odsReset, odsClear and odsHide

packages/common/core/src/form/interfaces/events.ts
skhamvon: added

aesteves60 (approved):

PR #321, opened 2023-12-18
feat(input): add form associated

aesteves60:
Add form associated on `input` component
Update to stencil 4.8.2 for the callback `formResetCallback`
Update `button` for support type reset

### DIFF
--- a/packages/components/input/package.json
+++ b/packages/components/input/package.json
@@ -44,7 +44,6 @@
   "devDependencies": {
     "@ovhcloud/ods-common-scripts": "16.5.0",
     "@ovhcloud/ods-common-testing": "16.5.0",
-    "@ovhcloud/ods-component-button": "16.5.0",
     "@ovhcloud/ods-stencil-dev": "16.5.0"
   }
 }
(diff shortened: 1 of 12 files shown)


packages/components/input/package.json
```diff
@@ -44,6 +44,7 @@
   "devDependencies": {
     "@ovhcloud/ods-common-scripts": "16.5.0",
     "@ovhcloud/ods-common-testing": "16.5.0",
+    "@ovhcloud/ods-component-button": "16.5.0",
```
dpellier, line 47: no needed anymore with the new arch, you can skip it already :)

packages/components/input/src/index.html
```diff
@@ -20,7 +20,24 @@
   </style>
 </head>
 <body style="margin: 0;">
-  <osds-input inline type="text" value="On Vous Héberge ?" disabled clearable></osds-input>
```
dpellier, line 23: Can you keep some test of input outside of a <form> element to ensure they can still be used?

dpellier (changes requested):

dpellier (approved):

astagnol (approved):

PR #328, opened 2023-12-26
feat(pagination): add odsPaginationItemPerPageChanged on itemPerPage change

dpellier:
Trigger a `odsPaginationItemPerPageChanged` event on `itemPerPage` select change, with the following payload:
```
current: 10 // new select value
currentPage: 1 // new page after change
totalPages: 6 // new number of pages after change
```

This replace the ods event triggered by the internal `ods-select` component


aesteves60 (changes requested):
Need to regenerate the doc

aesteves60 (approved):

PR #330, opened 2023-12-27
fix(theme): fixing fonts on Safari

manoncarbonnel:
Fixing fonts on Safari by:

- Re encoding font files in base64
- Spliting base64 output file to prevent linter error on huge file
- Removing now unused .stylelingignore files


aesteves60 (approved):

dpellier (changes requested):
font-weight seems to be wrong (see select example on Firefox which differ from the one in PROD)

astagnol (approved):

PR #337, opened 2024-01-10
fix(modal): prevent scroll when modal is open

skhamvon:
Prevent scroll when modal is opened

### DIFF
--- a/packages/components/src/modal/src/components/osds-modal/osds-modal.scss
+++ b/packages/components/src/modal/src/components/osds-modal/osds-modal.scss
@@ -35,7 +35,7 @@
   }
 
   .wrapper {
-    position: absolute;
+    position: fixed;
     z-index: 1;
     outline: none;
     border: none;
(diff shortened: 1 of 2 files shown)


dpellier (changes requested):
I'm a bit concern about deleting property of inetgrator element, it could have set the overflow value on purpose and we shouldn't remove it.
Can you try to add this behaviour through an ODS named class instead?

aesteves60 (approved):

dpellier (changes requested):
does not work on storybook
Capture d’écran 2024-01-12 à 16 00 14

dpellier (changes requested):
Ok initial render seems okay, but there is another issue regarding scrolling, the modal is positionned relatively of its trigger, if the trigger is scrolled, the modal deos not open centered.
To reproduce:
- go to the storybook modal page
- close the modal
- inspect the sandbox and add a long height to one of the container (ex: `<div id="root">`
- scroll to put the trigger button on the very top edge of the page
- click on the button
- the modal appear on the top edge, and partially hidden

PR #344, opened 2024-01-17
fix(modal): elements get truncated

Leotheluck:
Content:
- Fixing display problems within the modal that could lead to some elements getting shrunk.
- `max-height` of `OsdsModal`'s content has been raised to `50vh`.
- TS Linters have been fixed.

To review:
Check files `osds-modal.scss` and `osds-modal.size.scss` to review the **main** fix.
Check the other files to review **linters** fix.

To reproduce & compare the initial bug:
Add a lot of content inside an `OsdsModal` so that it becomes scrollable and add a few `OsdsInput` to see how they get truncated.

### DIFF
--- a/packages/components/src/modal/src/components/osds-modal/osds-modal.scss
+++ b/packages/components/src/modal/src/components/osds-modal/osds-modal.scss
@@ -50,13 +50,13 @@
     }
 
     .content {
-      max-height: 30vh;
+      max-height: 50vh;
       overflow-y: auto;
 
-      .slot {
-        display: flex;
-        flex-direction: column;
-      }
+      // .slot {
+      //   max-height: 30vh;
+      //   overflow-y: scroll;
+      // }
 
       .actions {
         display: flex;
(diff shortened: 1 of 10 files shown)


packages/components/src/modal/src/components/osds-modal/osds-modal.scss
```diff
@@ -50,13 +50,13 @@
...
     .content {
-      max-height: 30vh;
+      max-height: 50vh;
       overflow-y: auto;
 
-      .slot {
-        display: flex;
-        flex-direction: column;
-      }
+      // .slot {
```
dpellier, line 56: remove comment or explain why

packages/components/src/modal/src/components/osds-modal/osds-modal.spec.ts
```diff
@@ -38,52 +36,52 @@ describe('spec:osds-modal', () => {
 
   describe('attributes', () => {
     const config = {
-      instance: () => instance,
-      page: () => page,
-      root: () => page.root,
-      wait: () => page.waitForChanges(),
+      instance: (): OsdsModal => instance,
+      page: (): SpecPage => page,
+      root: (): HTMLElement | undefined => page.root,
```
dpellier, line 41: you can use `SpecPage['root']` directly

dpellier (changes requested):

dpellier (approved):

PR #345, opened 2024-01-17
refactor(textarea): apply common form logic

dpellier:


aesteves60 (approved):

packages/components/src/textarea/src/components/osds-textarea/constants/default-attributes.ts
astagnol, line 1: cols and rows default attributes no set?

astagnol (changes requested):
- inline, resizable, cols and rows controls in Storybook are not working
- contrasted control should be removed from Storybook

packages/components/src/textarea/src/components/osds-textarea/constants/default-attributes.ts
dpellier: They're optional, so unless we actually want a real default value, there is no need to set them.
And for what I see in the existing component, they were set a `undefined` by default

packages/components/src/textarea/src/components/osds-textarea/osds-textarea.tsx
```diff
@@ -1,289 +1,136 @@
...
     return (
-      <Host {...{
-        class: {
-          'ods-error': Boolean(this.controller.hasError()),
-        },
-        tabindex: textInputTabIndex,
-        hasFocus,
-      }}
-      >
+      <Host class={{ 'ods-error': this.internalError }}>
```
aesteves60, line 113: I think we need the tabindex here or the component will not be reflect on the host

packages/storybook/stories/controls.ts
```diff
@@ -0,0 +1,34 @@
...
+  CONTROL_CATEGORY.accessibility,
+];
+
+function orderControls(control: Record<string, any>) {
+  try {
+    return Object.entries(control)
+      .sort((a, b) => {
+        return orderedControlCategories.indexOf(a[ 1 ].category) - orderedControlCategories.indexOf(b[ 1 ].category);
+      })
+      .reduce((res, entry) => {
```
aesteves60, line 21:
```suggestion
      .reduce((res, { key, value }) => {
```

aesteves60 (commented):
You need to run yarn doc & remove the unused props like prefixValue or step

packages/storybook/stories/controls.ts
dpellier: entry is an array containing [<prop>, <value>] not an object

aesteves60 (approved):

astagnol (approved):

PR #346, opened 2024-01-17
fix(code): change icon contrasted state according to code

skhamvon:
Change icon contrasted state when code contrasted attribute is changed


dpellier (changes requested):
not sure it is really necessary to handle it, as this component as a free slot.
so the integrator should manage the `contrasted` attr directly on all the ods component he use.

PR #351, opened 2024-01-19
Feat/step 4 storybook

astagnol:
**Global**

- on component Demo, when control "contrasted" is set to true, background-color switch to dark tone
- add missing slots to Demo
- rework all Usage Guidelines (missing attributes, add slots, example that could be useful to our users and so on)
- add Canvas tag to all components usage (Preview being deprecated)
- disable Storybook shortcuts

**Fix**

- Broken url for logo (brand url) has been fixed
- Github logo that wasn't correctly displayed in prod should now be displayed as expected
- Component category (like Actions, Content and so on) that closed on refresh when you were on child page (Demo etc.) won't close anymore and you should stay where you were on refresh


dpellier (changes requested):
Few things I have noticed on Usage Guidelines pages:
- skeleton: last example (size) overflows horizontally
- modal: would be nice to have actual trigger on each case if possible, also why the color list here?
- search bar: loading example does not appear
- tile:
    - loading example does not appear
    - hoverable hollow does nothing
- select: openend example does not appear
- range: dual example does not appear
- quantity: using vanilla elements does not appear
- datepicker | password | phone number: typing in example trigger storybook shortcut
- form field: missing error example
- text: size overflows on small screen
- medium: image overflows on small screen
- menu: missing disabled example

dpellier (approved):

Leotheluck (approved):

PR #353, opened 2024-01-22
fix(select): select automatically option if there is only one

skhamvon:


dpellier (changes requested):
Regarding UX, should we select the first option only if the field is required?
As empty value may also be a valid choice

PR #354, opened 2024-01-23
feat(pagination): add defaultItemsPerPage attribute

Leotheluck:
This new feature will allow the users to select the default number of items displayed per page in the `OsdsPagination` component.

Content:
- Added new attribute `defaultItemsPerPage` to `OsdsPagination`.
- Added new type `ODS_PAGINATION_PER_PAGE`.

To review:
Check if you are able to add a `defaultItemsPerPage` attribute to your `OsdsPagination` component, and if it applies correctly.


packages/components/src/pagination/src/components/osds-pagination/constants/pagination-per-page.ts
```diff
@@ -1,9 +1,14 @@
 const ODS_PAGINATION_PER_PAGE_MIN = 10;
 const ODS_PAGINATION_PER_PAGE_MAX = 300;
 const ODS_PAGINATION_PER_PAGE_OPTIONS = Object.freeze([ODS_PAGINATION_PER_PAGE_MIN, 25, 50, 100, ODS_PAGINATION_PER_PAGE_MAX]);
+type ODS_PAGINATION_PER_PAGE = typeof ODS_PAGINATION_PER_PAGE_OPTIONS[number];
```
dpellier, line 4: better to transform the above as an enum (like other constants in the project)

packages/components/src/pagination/src/components/osds-pagination/osds-pagination.tsx
```diff
@@ -24,16 +25,17 @@ export class OsdsPagination implements OdsPaginationAttribute, OdsPaginationEven
...
-
   @Prop({ mutable: true, reflect: true }) current: number = DEFAULT_ATTRIBUTE.current;
   @Prop({ reflect: true }) totalItems?: number;
+  @Prop({ reflect: true }) defaultItemsPerPage: ODS_PAGINATION_PER_PAGE = DEFAULT_ATTRIBUTE.defaultItemsPerPage;
   @Prop({ reflect: true }) totalPages: number = DEFAULT_ATTRIBUTE.totalPages;
   @Prop({ mutable: true , reflect: true }) disabled: boolean = DEFAULT_ATTRIBUTE.disabled;
   @Prop({ reflect: true }) labelTooltipPrevious: string = DEFAULT_ATTRIBUTE.labelTooltipPrevious;
   @Prop({ reflect: true }) labelTooltipNext: string = DEFAULT_ATTRIBUTE.labelTooltipNext;
 
+  @State() itemPerPage = this.defaultItemsPerPage ? this.defaultItemsPerPage : ODS_PAGINATION_PER_PAGE_MIN;
```
dpellier, line 36: `= this.defaultItemsPerPage || ODS_PAGINATION_PER_PAGE_MIN;` ?

dpellier (changes requested):

astagnol (changes requested):
- New attribute `defaultItemsPerPage` is not displayed in Storybook Specifications
- We should add that new attribute to controls in Storybook (Total Items page)
- Please add an integration example for this attribute to Usage Guidelines

dpellier (changes requested):
seems KO on demo page
Capture d’écran 2024-01-25 à 09 13 38

dpellier (changes requested):
New issue dtected on storybook.
As the `defaultItemsPerPage` control field is a number input, we can put a value that is not handled by the component (ex: set 26 and reload => empty select)

2 thing to do here:
- manage this edge case on component side to default to min value if the value set is incorrect (+ tests)
- update the story page with a select control instead
```
control: 'select',
options: ODS_PAGINATION_PER_PAGE_OPTIONS,
```

dpellier (approved):

packages/components/src/pagination/documentation/usage-guidelines/usage.mdx
```diff
@@ -17,10 +17,10 @@ import Methods from './methods/usage.methods.mdx';
 
 ### Total Items
```
astagnol, line 18: I think we should mention "Default items per page"here in the title with "Total items" since they work together

packages/storybook/stories/components/pagination/3_demo.stories.ts
astagnol, line 1: When I pick up an option from defaultItemsPerPage select in Controls, it does not apply to the component in the Story. 
For example, I've set total items  to '100' and picked up '100' from defaultItemsPerPage select. Nothing happens until I refresh the page

astagnol (changes requested):

packages/storybook/stories/components/pagination/3_demo.stories.ts
dpellier: default-like props are used only on initial render, so if you change the value, you'll have to reload to see the change.

astagnol: Alright, thanks. 
Then, I take note that maybe I should specify this in the control description so our users won't be surprised

astagnol (approved):

PR #355, opened 2024-01-23
fix(toggle): auto adapt slot container when content is too long

skhamvon:
Adapt automatically slot container  width when content is too long


dpellier (changes requested):
Break the toggle size when putting large content

Capture d’écran 2024-01-24 à 09 17 08

dpellier (approved):

PR #358, opened 2024-01-25
Refactor/phone form

aesteves60:


packages/common/core/src/form/method.controller.ts
```diff
@@ -64,6 +63,10 @@ class OdsCommonFieldMethodController<T extends OdsCommonFieldAttribute & OdsComm
     }
   }
 
+  async setTabindex(value: number): Promise<void> {
```
dpellier, line 66: has been removed for all other, do we need to add it again for this one to work?

dpellier (changes requested):

astagnol (approved):
OK but we need to discuss about setTabindex

PR #359, opened 2024-01-25
fix(tile): end slot position when there is no start slot

skhamvon:
Fix end slot position when there is no start slot in the tile


astagnol (changes requested):
Capture d’écran 2024-01-30 à 11 00 14

Issue still persists on Chromium

astagnol (approved):

dpellier (approved):

PR #362, opened 2024-01-26
fix(input): fix icon size

skhamvon:
Fix input icon size using his own size instead of one setted for input component


astagnol (changes requested):
Capture d’écran 2024-01-30 à 10 50 47

On safari, native arrows are displayed (I've checked master and it is OK)

dpellier:
@astagnol seems not related to this PR as I see the arrow on prod documentation also.
Should be treated in another branch

PR #368, opened 2024-02-06
fix(modal): closeModal event is called twice on unmount

skhamvon:
closeModal event is called twice on unmount

### DIFF
--- a/packages/components/src/modal/src/components/osds-modal/osds-modal.tsx
+++ b/packages/components/src/modal/src/components/osds-modal/osds-modal.tsx
@@ -87,7 +87,8 @@ export class OsdsModal implements OdsModalAttribute, OdsModalMethod, OdsModalEve
   }
 
   disconnectedCallback(): void {
-    this.close();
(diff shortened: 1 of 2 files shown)


packages/components/src/modal/src/components/osds-modal/osds-modal.tsx
```diff
@@ -87,7 +87,8 @@ export class OsdsModal implements OdsModalAttribute, OdsModalMethod, OdsModalEve
   }
 
   disconnectedCallback(): void {
-    this.close();
+    document.body.style.removeProperty('overflow');
+    this.masked = true;
```
dpellier, line 91: put those in a dedicated function and call it from here and close `method`

dpellier (changes requested):

packages/components/src/modal/src/components/osds-modal/osds-modal.tsx
skhamvon: Done

dpellier (approved):

PR #373, opened 2024-02-14
Refactor/datepicker2

aesteves60:


packages/components/src/datepicker/src/components/osds-datepicker/osds-datepicker.tsx
```diff
@@ -33,17 +35,24 @@ Object.assign(Datepicker.locales, nl);
 Object.assign(Datepicker.locales, pl);
 Object.assign(Datepicker.locales, pt);
 
+// TODO prevent onBlur when tabbing from input to datepicker? => quid clearable?
+// TODO test on ods-form
+//  - issue with required
+//  - issue with reset => OK with form / OK with osds-form
+// TODO test on react example => OK
+// TODO shall we trigger value event when value is not a valid date? => see pattern
```
dpellier, line 43: all resolved?

astagnol (approved):

packages/components/src/input/src/components/osds-input/osds-input.tsx
```diff
@@ -153,7 +153,7 @@ export class OsdsInput implements OdsInputAttribute, OdsInputEvent, OdsInputMeth
 
     return (
       <Host
-        class={{ 'ods-error': this.internalError }}
```
dpellier, line 156: `class={{ 'ods-error': this.error || this.internalError }}` ?
would be nice to avoid as much as possible to allow mutable attr

dpellier (changes requested):

packages/components/src/input/src/components/osds-input/osds-input.tsx
aesteves60: I remove the input edit for add it in the v18 PR 
https://github.com/ovh/design-system/pull/293/commits/a537288c5cc5808db9682ae7e11c2331b14e4688

Leotheluck (changes requested):
- Elements inside the Datepicker popup should be blue (currently grey)
- If you open the Datepicker popup by clicking on the calendar icon, it becomes impossible to blur the popup by clicking outside (which works if we open it by clicking on the input)

---

PR #376, opened 2024-02-14
fix(pagination): hide navigation when totalitems lower itemsPerPage

skhamvon:
Hide navigation of pagination when totalItems is lower than itemsPerPage


manoncarbonnel (approved):

dpellier (changes requested):
waiting for spec update, there are some weird behavior because of the select values

aesteves60 (approved):

dpellier (changes requested):
Select should be visible all the time
Capture d’écran 2024-02-20 à 10 25 28
Capture d’écran 2024-02-20 à 10 25 34

astagnol (changes requested):
Capture d’écran 2024-02-20 à 16 23 18

Arrows and one page (1) buttons should be displayed when page amount is less than 2

dpellier (approved):

astagnol (approved):

PR #377, opened 2024-02-14
fix(textarea): change style to be like osds-input

skhamvon:
Change style of textarea to be more like input


dpellier (changes requested):
`disabled` style still differs, not the right icon and change color on click

Leotheluck (approved):

dpellier (approved):

astagnol (approved):

PR #378, opened 2024-02-15
fix(datepicker): set days to 2 letters

skhamvon:
Set days name to 2 letters


Leotheluck (approved):

aesteves60 (approved):

dpellier (changes requested):
@astagnol what is the expected rendering here to ensure consistent day names for all locales?
For example, `fr` and `pl` render differently than `en`
Do we want 2 letters everywhere, with same case?

@dpellier This issue is more about displaying "en" locale as it should be with the library than set days to 2 letters. Since the library for "en" locale is "Su", "Mo", "Tu" and so on, we should display this when initializing the component.

We do not limit to 2 letters everywhere, some locales from the library are going to display 3 letters, for example.

dpellier (approved):

PR #382, opened 2024-02-16
fix(input): handler key space on icon button

aesteves60:


dpellier (approved):

astagnol (changes requested):
- On Safari v16.6, cannot focus on the Icon button
- On Safari v17, can focus on the Icon button but no style is applied (no border like on firefox)

aesteves60:
In BrowserStack, it's focusable with the good design in safari 16.5 & 17.3.
I can't try on 16.6

astagnol (approved):

PR #383, opened 2024-02-16
fix(tile): fix style of checking state in safari

skhamvon:


dpellier (changes requested):
to check on safari, when `checking` is true, the blink shows the white border rectangle

astagnol (changes requested):
Capture d’écran 2024-02-19 à 10 16 47
When checking is true, tile container border is visible. OK on other browsers

PR #385, opened 2024-02-21
Feat/input pattern

aesteves60:


dpellier (changes requested):
Strange behaviour on storybook when removing pattern value:
- go to default defaut
- add pattern (ex: `\d*`)
- type numbers => all good
- delete pattern value to empty string
==> input is always on error

packages/storybook/stories/components/datepicker/3_demo.stories.ts
```diff
@@ -80,6 +80,10 @@ const storyParams = {
     category: 'Development',
     defaultValue: true,
   },
+  pattern: {
```
dpellier, line 83: datepicker should not allow a custom pattern, it should match the Prop `format`

dpellier (changes requested):

dpellier (approved):

PR #386, opened 2024-02-22
Fix/clipboard tooltip

aesteves60:
Not display tooptip when slot `success-message` or `error-message` is not defined


dpellier (changes requested):
seems to still pop:
Capture d’écran 2024-02-22 à 14 18 07

aesteves60:
The bug with the tooltip empty is du to storybook, who add html comment !
I have this content in the slot
image

dpellier:
WDYT about checking the actual emptiness of slot using DOMParser ?
```
function getTextContent(html: string) {
    return new DOMParser()
        .parseFromString(html, "text/html")
        .documentElement.textContent;
}
```

dpellier (changes requested):
add some unit tests for each getTextContent possible values

dpellier (approved):

PR #388, opened 2024-02-23
feat(timepicker): add new component

skhamvon:
Add timepicker component


dpellier (changes requested):
Issue on default demo on storybook when:
- changing current timezone => empty select
- toggling error => nothing happens

Capture d’écran 2024-02-26 à 09 08 22

astagnol (changes requested):
When you change something in the controls (for example disabled from false to true), the field is cleared. (this is not happening on Input component of type time)

dpellier (changes requested):
Same issue a basic input about clearable button still focusable when disabled
(https://github.com/ovh/design-system/pull/293#pullrequestreview-1913581661)
waiting for resolution on v18 branch, then rebase and retest here

Leotheluck (approved):

PR #390, opened 2024-02-28
Fix/input focus on clear

aesteves60:
Focus on input when clear
Add focusable icon according to this PR https://github.com/ovh/design-system/pull/382 This PR was merge in V18

### DIFF
--- a/packages/components/src/input/src/components/osds-input/core/controller.spec.ts
+++ b/packages/components/src/input/src/components/osds-input/core/controller.spec.ts
@@ -59,6 +59,26 @@ describe('spec:ods-input-controller', () => {
       });
     });
 
+    describe('methods:handleKeySpace', () => {
+      it('should call the callback', () => {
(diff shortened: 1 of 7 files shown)


dpellier (changes requested):
Back tabbing does not work properly.
- go to the demo
- set clearable true
- focus input
  - tab -> button focused => ok
  - shift+tab -> focus is not on the input => ko, input should be focused again

(also does not work on Safari)

astagnol (changes requested):
shift + tab doesn't work properly

But in Safari v17.3.1:
- set clearable true
- focus input
- tab -> button focused => ok for me but no style is applied while focused

But there was the same issue here and it has been fixed (and merged in refactor/input-attrs): https://github.com/ovh/design-system/pull/382 (check my resolved comment)

dpellier (approved):
fix e2e then it looks good

Leotheluck (approved):

astagnol (approved):

PR #391, opened 2024-03-06
fix(clipboard): fix tooltip close delay

manoncarbonnel:
Adding a delay to close the surface tooltip and e2e test


packages/components/src/clipboard/src/components/osds-clipboard/core/controller.ts
```diff
@@ -42,6 +42,8 @@ class OdsClipboardController {
     }
     this.component.surfaceMessage = html;
     this.component.surface.opened = !this.component.surface.opened;
+
+    setTimeout(() => this.closeSurface(), 3000);
```
dpellier, line 46: Could be nice to also add a debounce on the `handlerClick` method, to avoid stacking timeout call on erratic click behavior.
There is a `odsDebounce` method available, you can see a usage example on the tooltip component

dpellier (changes requested):

PR #392, opened 2024-03-12
feat(doc): change generation doc

aesteves60:
Adapt generation doc with the new arch 
Script generation change the file `spec.md`.
- Now the list and not an array.
- Prop from stencil with the decorator `@Prop`
- Event from stencil with the decorator `@Event`
- Method from stencil with the decorator `@Method`
- Add enum
- Remove the doc from the interface attributes, events & methods
Change script generate component with the correct script `doc`

### DIFF
--- a/packages/ods/scripts/typedoc-plugin-decorator.js
+++ b/packages/ods/scripts/typedoc-plugin-decorator.js
@@ -1,0 +1,49 @@
+const td = require("typedoc");
+const ts = td.TypeScript;
+
+/** @param {td.Application} app */
+exports.load = function (app) {
+    // Add decorator info to reflections
+    // if you need parameters, you need app.converter.on(td.Converter.EVENT_CREATE_PARAMETER)
+    app.converter.on(td.Converter.EVENT_CREATE_DECLARATION, addDecoratorInfo);
+
+    // Add decorator info to serialized json
+    app.serializer.addSerializer({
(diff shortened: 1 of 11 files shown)


packages/ods/scripts/typedoc-plugin-decorator.js
```diff
@@ -0,0 +1,49 @@
+const td = require("typedoc");
```
dpellier, line 1: apply `editorconfig` on this file (spacing) + replace `"` by `'`

packages/ods/src/components/spinner/src/components/ods-spinner/ods-spinner.tsx
```diff
@@ -1,7 +1,7 @@
 import type { OdsSpinnerColor } from '../../constants/spinner-color';
 import type { OdsSpinnerSize } from '../../constants/spinner-size';
 import type { FunctionalComponent } from '@stencil/core';
-import { Component, Host, Prop, h } from '@stencil/core';
+import { Component, Event, EventEmitter, Host, Method, Prop, h } from '@stencil/core';
```
dpellier, line 4: `Spinner` is almost done, use only `text` to add test stuff

dpellier (changes requested):

dpellier (approved):

PR #401, opened 2024-03-18
fix(menu): force min height in style instead of CDK

skhamvon:
Fix bottom margin when there is one element by setting min height in menu style instead of automatically setted by CDK


dpellier (changes requested):
Please add a comment explaining why you had to rley on important here

skhamvon:
The use of !important is necessary because CDK is setting a min-height inline rule.
It is the only way to override it without making big modifications in the CDK system

feoche (approved):

astagnol (approved):

dpellier (approved):

Leotheluck (approved):

PR #403, opened 2024-03-19
fix(button): remove margin & padding of host

skhamvon:
Remove margin & padding to prevent click event outside of the component

### DIFF
--- a/packages/components/src/button/src/components/osds-button/core/controller.ts
+++ b/packages/components/src/button/src/components/osds-button/core/controller.ts
@@ -52,9 +52,7 @@ class OdsButtonController {
    * Handle Click and KeyPress Event on the button
    */
   handleClick(event: MouseEvent): void {
-    if (!this.component.disabled) {
-      this.submitForm(event);
-    }
+    this.submitForm(event);
   }
 
   handleKey(event: KeyboardEvent): void {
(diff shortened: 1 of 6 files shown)


dpellier (changes requested):
I've tried the React example and was abli to click the disabled button on Chromium and Safari

aesteves60 (changes requested):
We need a test that has failed before the fix

astagnol (changes requested):
Same here, I am able to click when disabled on several browsers (react example):
Firefox (123.0.1) -> can click just a few pixels outside the disabled button (but ok on the border)
Safari (Version 17.3.1 ) -> same
Chromium (Version 112.0.5615.49) -> can click on the disabled button (button, border, a few pixels outside of it)

Leotheluck (approved):

aesteves60 (approved):

dpellier (changes requested):
Still possible to call an attached click event listener on chromium.

What you can do, is add an inert prop on the host, like:
```
<Host ...
        inert: this.disabled
```

dpellier:
There is an issue with Chromium and `slot` content in a `disabled` element (see: https://issues.chromium.org/issues/40679491#comment13)
Issue is reproductible on other framework too.
A workaround could be to "silence" the slot using something like:
Capture d’écran 2024-04-02 à 15 05 55

Please give it a try following the whole process:
- reproduce the issue on your local Chromium
- apply the fix
- test on Chromium / FF / Safari (/ Edge)

feoche (approved):

PR #410, opened 2024-03-27
feat(autocomplete): create component

Leotheluck:
## Description

`OsdsAutocomplete` is a new component that allows users to type into an input that will open a dropdown displaying a list of selectable options from which the user can choose from.

> Keep in mind the `OsdsAutocomplete` component is not meant to implement the logic of sorting or filtering the elements that we put inside it. Example of one possible filtering logic is to be found in the `index.html` but is only for testing purpose.

## Content

- New component `OsdsAutocomplete`
- Documentation
- Storybook stories

## To review

Check if the component works properly in multiple environments and situations. Also make sure that documentation is up to the requirements. Look for any potential mistakes or improvements we could address.

## To be added

- Loading state
- BEM SCSS & remove !important (there was an issue with linters being investigated)
- Add a "With logic" page to Storybook

### DIFF
--- a/packages/components/src/autocomplete/documentation/usage-guidelines/usage.mdx
+++ b/packages/components/src/autocomplete/documentation/usage-guidelines/usage.mdx
@@ -1,3 +1,4 @@
+import WithLogic from './properties/usage.logic.mdx';
 import ClearableProperty from './properties/usage.clearable.mdx';
 import DisabledProperty from './properties/usage.disabled.mdx';
 import ErrorProperty from './properties/usage.error.mdx';
@@ -30,6 +31,9 @@ import { Canvas } from '@storybook/addon-docs';
     <osds-select-option value="EN">Hello</osds-select-option>
 </osds-autocomplete>
 ```
+
+### Autocompletion Logic
+<WithLogic />
 
 ### Clearable
 <ClearableProperty />
--- a/packages/components/src/autocomplete/documentation/usage-guidelines/usage.mdx
+++ b/packages/components/src/autocomplete/documentation/usage-guidelines/usage.mdx
@@ -1,0 +1,56 @@
+import ClearableProperty from './properties/usage.clearable.mdx';
+import DisabledProperty from './properties/usage.disabled.mdx';
+import ErrorProperty from './properties/usage.error.mdx';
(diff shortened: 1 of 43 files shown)


packages/components/src/autocomplete/src/globals.ts
```diff
@@ -0,0 +1,12 @@
+/**
+ * Import here all the external ODS component that you need to run the current component
+ * when running dev server (yarn start) or e2e tests
+ *
+ * ex:
+ *   import '../../text/src';
+ */
+import '../../content-addon/src';
```
dpellier, line 8: ?

packages/components/src/autocomplete/documentation/usage-guidelines/usage.mdx
```diff
@@ -0,0 +1,60 @@
...
+<DisabledProperty />
+
+### Error
+<ErrorProperty />
+
+### Icon
+<IconProperty />
+
+### Inline
+<InlineProperty />
```
dpellier, line 51: Inline is set twice

dpellier (changes requested):
Issue when doing the following:
- on Demo page, click on input
- click on `Bonjour`
- click on the clearable icon
- click on input to reopen the list
- click on `Bonjour`
==> Empty inpuy

astagnol (changes requested):
**Global**
There is no 'no results found' text to display for the Autocomplete component but it would be great to show how to integrate this in the example with logic

**Keyboard navigation**
We should not have to 'Tab' twice to be able to type in the field (careful with Safari, currently there is no focus at all on 'Tab', it exits the field)

When the dropdown is displayed, 'Tab' should enable focus on the next element but not the options:
- if the user has moved forward through the list of options with 'arrow', the focused option = selected value & the dropdown is closed on 'Tab'. Since an option is selected, we should focus on clearable Icon.
- if no option is focused yet, the dropdown is closed & the next focusable element is focused on 'Tab'

When the user selects an option on 'Enter', dropdown is closed & the field is in state 'focus'.

**Mouse**
The dropdown is closed & the field is in state 'focus' when the user clicks on an option

dpellier:
Will be reimplemented on v18 when ready

PR #412, opened 2024-03-28
feat(autocomplete): loading state

Leotheluck:
## Description

`OsdsAutocomplete` is a new component that allows users to type into an input that will open a dropdown displaying a list of selectable options from which the user can choose from.

This addition allows users to add an `isLoading` attribute to the component.

## Content

- `isLoading` will display an `OsdsSpinner` inside the Autocomplete component instead of the slot for the options.

## To review

Check if the `isLoading` attribute works and reacts properly with the correct styling.

### DIFF
--- a/packages/components/src/autocomplete/src/components/osds-autocomplete/osds-autocomplete.scss
+++ b/packages/components/src/autocomplete/src/components/osds-autocomplete/osds-autocomplete.scss
@@ -39,6 +39,14 @@
   animation: appear 0.2s ease-in-out;
 }
 
+:host(.ods-autocomplete) .anchor .overlay .loading {
+  display: flex;
+  align-items: center;
+  justify-content: center;
+  padding: var(--ods-size-03);
+  height: var(--ods-size-09);
+}
+
 :host(.ods-autocomplete[opened]) {
   .anchor .overlay {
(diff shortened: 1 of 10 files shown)


packages/components/src/autocomplete/src/components/osds-autocomplete/osds-autocomplete.scss
```diff
@@ -39,6 +39,14 @@
   animation: appear 0.2s ease-in-out;
 }
 
+:host(.ods-autocomplete) .anchor .overlay .loading {
```
aesteves60, line 42: Can you use the nested implementation ?

aesteves60 (changes requested):

packages/components/src/autocomplete/src/components/osds-autocomplete/osds-autocomplete.scss
Leotheluck: Yes it's a topic I wanted to cover once this one would be merged, I had issues with eslint using the nested, let's cover it in the other PR?

dpellier (approved):

aesteves60 (approved):

PR #415, opened 2024-04-04
Refactor/ods button

dpellier:
(does not contains boolean attribute normalization, it will come on another dedicated PR)


packages/ods/src/components/button/src/components/ods-button/ods-button.tsx
```diff
@@ -0,0 +1,70 @@
...
+            <ods-spinner
+              class="ods-button__button__spinner"
+              color={ ODS_SPINNER_COLOR.neutral }
+              size={ ODS_SPINNER_SIZE.sm }>
+            </ods-spinner>
+          }
+
+          {
+            !!this.icon &&
+            <ods-icon
```
aesteves60, line 59: I think we also need a part on the icon for customized it.

packages/ods/src/style/_button.scss
```diff
@@ -0,0 +1,153 @@
...
+    @if $variant == 'ghost' {
+      background-color: inherit;
+    } @else if $variant == 'outline' {
+      border-color: var(--ods-color-neutral-300);
+    }
+  }
+
+  &:not(:disabled) {
+    &:focus {
+      outline: solid 2px var(--ods-color-primary-700);
```
aesteves60, line 27:
```suggestion
       outline: var(--ods-outline-color) var(--ods-outline-style) var(--ods-outline-width);
       outline-offset: var(--ods-outline-offset);
```

aesteves60 (changes requested):

packages/ods/src/components/button/src/components/ods-button/ods-button.tsx
dpellier: Not use there is an actual use case for this.
Color and size will always match the label (now that we use a font for the icon)

astagnol (approved):
Global:
- For default button in primary color on hover, border color should be primary-700 instead of 500.
- Same goes for critical color on hover, border color should be critical-700.
- For default button in primary color on pressed, background & border colors should be primary-800
- Disabled and loading are missing in controls for Demo

Safari 17.3.1:
Need to "tab" twice to get expected focus style. Here's the style on first "tab":
Capture d’écran 2024-04-08 à 16 23 26

Firefox v124.0.1 & Chromium v112.0.5615.49:
Focus style is displayed on click and it persists until I click somewhere else:
Capture d’écran 2024-04-08 à 16 26 01

astagnol (changes requested):
Check previous comment

astagnol (approved):

PR #419, opened 2024-04-11
feat(tag): implement component

dpellier:


packages/ods/src/style/_tag.scss
```diff
@@ -0,0 +1,178 @@
+@import './focus';
+
+$ods-tag-height-sm: 16px;
+$ods-tag-height-md: 20px;
+$ods-tag-height-lg: 28px;
+
+@mixin ods-tag() {
```
aesteves60, line 7: I think it miss the font family

aesteves60 (changes requested):

packages/storybook/stories/components/tag/tag.stories.ts
```diff
@@ -0,0 +1,175 @@
...
+        defaultValue: { summary: 'ø' },
+        type: { summary: 'string' }
+      },
+      control: 'text',
+      description: 'Set a custom style properties. Example: "background-color: #008000;"',
+    },
+    icon: {
+      table: {
+        category: CONTROL_CATEGORY.design,
+        type: { summary: ODS_ICON_NAMES }
```
aesteves60, line 55: For the icon the list is very long, maybe a link to the icon doc should be better

aesteves60 (changes requested):

astagnol (approved):

astagnol (changes requested):
- Close Icon in Tag md size should be 1rem instead of 0.875 and the one in lg size should be 1.25rem instead of 1rem.

- Tag in small spaces:
Capture d’écran 2024-04-12 à 08 57 52

aesteves60 (approved):

PR #434, opened 2024-04-18
feat(medium): implement component

dpellier:


astagnol (approved):
Maybe we should use an image not so large for the documentation?

aesteves60 (approved):

PR #439, opened 2024-04-26
Refactor/ods quantity

aesteves60:
Refactor component Quantity

- Hover on button the border color change but not the icon color. should be change ?


feoche (changes requested):
### Documentation
- [x] Remove `placeholder` example
- [x] Add default value on `readonly` example

### Migration doc
- [x] Typo on `araiLabelledby` 

### Demo
- [x] Is in error and disabled at start
- [x] Remove `defaultValue` control
- [ ] Setting `max` value to inferior value of current does not disable the button
- [ ] Same for `min`
- [x] Remove `placeholder` control

### Style
- [x] Left and right border of Input should not be visible
- [x] Text is not centered
- [x] Text can be entered inside Input
- [x] Button does not have a hover effect
- [x] Button text color does not change in hover/focus state
- [x] Outline : prefer `z-index: 1` on focus-visible 
- [x] Readonly : Buttons should not be clickable
- [x] Error style overrides disabled style
- [x] 4 characters can't be visible
- [ ] `min/max` are not checked when blurring Input
- [x] Outlines are not the same when in Error state

feoche (changes requested):
### Migration
- [x] Change "Has been add." → "New attribute (optional)"
- [x] Typo "Has been remove." → "Has been removed."
- [x] Typo "automaticly" → "automatically"

### Style
- [x] Set padding to `4px` on Input

packages/ods/src/components/quantity/src/components/ods-quantity/ods-quantity.tsx
```diff
@@ -0,0 +1,146 @@
...
+  render(): FunctionalComponent {
+    return (
+      <Host class="ods-quantity">
+        <ods-button
+          class={{
+            'ods-quantity__button': true,
+            'ods-quantity__button--readonly': this.isReadonly,
+          }}
+          color={ this.hasError ? ODS_BUTTON_COLOR.critical : ODS_BUTTON_COLOR.primary }
+          exportparts="button:button_minus"
```
dpellier, line 99: `button-minus` (classic case for css)

packages/ods/src/components/quantity/src/components/ods-quantity/ods-quantity.tsx
```diff
@@ -0,0 +1,146 @@
...
+          isDisabled={ isMinusButtonDisabled(this.isDisabled, this.value, this.min) }
+          icon={ ODS_ICON_NAME.minus }
+          label=""
+          onClick={ () => this.decrement() }
+          onFocus={ () => this.odsFocus.emit() }
```
dpellier, line 104: shall we really emit here ? on form element focus and blur should be emitted by the input(s) only

packages/ods/src/components/quantity/src/components/ods-quantity/ods-quantity.tsx
```diff
@@ -0,0 +1,146 @@
...
+          hasError={ this.hasError }
+          isDisabled={ this.isDisabled }
+          isReadonly={ this.isReadonly }
+          isRequired={ this.isRequired }
+          onOdsChange={ (event: OdsInputValueChangeEvent) => this.onOdsChange(event) }
+          max={ this.max }
+          min={ this.min }
+          name={ this.name }
+          placeholder={ this.placeholder }
+          ref={ (el?: unknown) => this.odsInput = el as OdsInput }
```
dpellier, line 123: use `HTMLOdsInputElement` instead? (by changing also odsInput type to HTMLOdsInputElement

packages/ods/src/components/quantity/src/components/ods-quantity/ods-quantity.tsx
```diff
@@ -0,0 +1,146 @@
...
+          type={ ODS_INPUT_TYPE.number }
+          value={ this.value }>
+        </ods-input>
+        <ods-button
+          class={{
+            'ods-quantity__button': true,
+            'ods-quantity__button--readonly': this.isReadonly,
+          }}
+          color={ this.hasError ? ODS_BUTTON_COLOR.critical : ODS_BUTTON_COLOR.primary }
+          exportparts="button:button_add"
```
dpellier, line 134: `button-plus`

packages/ods/src/components/quantity/src/controller/ods-quantity.ts
```diff
@@ -0,0 +1,17 @@
+function isAddButtonDisabled(isDisabled: boolean, value: number | null, max?: number): boolean {
+  return isDisabled || (max !== undefined && max === value);
```
dpellier, line 2: this enabled button if value is bigger than the max (which is doable by typing directly the value in the input)
`return isDisabled || (max !== undefined && value >= max);`

dpellier: Add tests accordingly

packages/ods/src/components/quantity/src/controller/ods-quantity.ts
```diff
@@ -0,0 +1,17 @@
...
+}
+
+function isMinusButtonDisabled(isDisabled: boolean, value: number | null, min?: number): boolean {
+  return isDisabled || (min !== undefined && min === value);
```
dpellier, line 6: same issue here when typing value less than min, button get enabled again

packages/ods/src/components/quantity/tests/behaviour/ods-quantity.e2e.ts
```diff
@@ -0,0 +1,141 @@
...
+    buttonAdd = await page.find('ods-quantity >>> [exportparts="button:button_add"]');
+
+    await page.waitForChanges();
+  }
+
+  beforeEach(jest.clearAllMocks);
+
+  describe('method:clear', () => {
+    it('should receive odsClear event', async() => {
+      await setup(`<ods-quantity value="${0}"></ods-quantity>`);
```
dpellier, line 27: why the `${}` everywhere?

packages/ods/src/components/quantity/tests/behaviour/ods-quantity.e2e.ts
```diff
@@ -0,0 +1,141 @@
...
+    it('should not increment value with is-readonly', async() => {
+      await setup(`<ods-quantity value="${0}" is-readonly></ods-quantity>`);
+      const odsChangeSpy = await page.spyOnEvent('odsChange');
+      await buttonAdd.click();
+      await page.waitForChanges();
+      expect(await el.getProperty('value')).toBe(0);
+      expect(odsChangeSpy).toHaveReceivedEventTimes(0);
+    });
+
+    it('should not decrement value with max equal to value', async() => {
```
dpellier, line 132: add test if value is more than max

packages/ods/src/components/quantity/tests/behaviour/ods-quantity.e2e.ts
```diff
@@ -0,0 +1,141 @@
...
+    it('should not decrement value with is-readonly', async() => {
+      await setup(`<ods-quantity value="${0}" is-readonly></ods-quantity>`);
+      const odsChangeSpy = await page.spyOnEvent('odsChange');
+      await buttonMinus.click();
+      await page.waitForChanges();
+      expect(await el.getProperty('value')).toBe(0);
+      expect(odsChangeSpy).toHaveReceivedEventTimes(0);
+    });
+
+    it('should not decrement value with min equal to value', async() => {
```
dpellier, line 85: add test if value is less than min

packages/ods/src/components/quantity/tests/rendering/ods-quantity.e2e.ts
```diff
@@ -0,0 +1,132 @@
...
+      </form>`);
+      await buttonAdd.click();
+      const submitButton = await page.find('button[type="submit"]');
+      await submitButton.click();
+      await page.waitForNetworkIdle();
+      const url = new URL(page.url());
+      expect(url.searchParams.get('ods-quantity')).toBe('1');
+    });
+
+    it('should get form data with button type submit after increment quantity', async() => {
```
dpellier, line 103: you inverted this test with the one above (increment click)

packages/ods/src/components/quantity/tests/rendering/ods-quantity.e2e.ts
```diff
@@ -0,0 +1,132 @@
...
+    it('should render the web component with has-error', async() => {
+      await setup('<ods-quantity has-error></ods-quantity>');
+
+      expect(input.getAttribute('has-error')).toBe('');
+      expect(buttonMinus.getAttribute('color')).toBe(ODS_BUTTON_COLOR.critical);
+      expect(buttonAdd.getAttribute('color')).toBe(ODS_BUTTON_COLOR.critical);
+    });
+  });
+
+  describe('Form', () => {
```
dpellier, line 88: probably more a behaviour kind of test, or either a specific `form` test category?

packages/ods/src/components/quantity/tests/rendering/ods-quantity.e2e.ts
```diff
@@ -0,0 +1,132 @@
...
+      const submitButton = await page.find('button[type="submit"]');
+      await submitButton.click();
+      await page.waitForNetworkIdle();
+      const url = new URL(page.url());
+      expect(url.searchParams.get('ods-quantity')).toBe('0');
+    });
+
+    it('should reset form with button type reset', async() => {
+      await setup(`<form method="get">
+        <ods-quantity name="ods-quantity" value="On Vous Heberge ?"></ods-quantity>
```
dpellier, line 118: numerical value

dpellier (changes requested):

packages/ods/src/components/quantity/src/components/ods-quantity/ods-quantity.tsx
aesteves60: In the mozilla doc the name is like this https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/exportparts#exposing_mapped_parts

packages/ods/src/components/quantity/src/components/ods-quantity/ods-quantity.tsx
aesteves60: Don't work on the CI test

packages/ods/src/components/quantity/src/controller/ods-quantity.ts
aesteves60: It's because of the input value, it's like type text in the input, the event `onInput` is not trigger and the input is not valid

packages/ods/src/components/quantity/src/components/ods-quantity/ods-quantity.tsx
dpellier: it's a double `_` on the doc, probably from a BEM exemple, part are used on css side, so generic case would be with a `-`

packages/ods/src/components/quantity/tests/navigation/ods-quantity.e2e.ts
```diff
@@ -28,6 +28,7 @@ describe('ods-quantity navigation', () => {
       await setup('<ods-quantity></ods-quantity>');
       const odsFocusSpy = await page.spyOnEvent('odsFocus');
       await page.keyboard.press('Tab');
+      await page.keyboard.press('Tab');
```
dpellier, line 31: why 2 tab?

packages/storybook/stories/components/quantity/migration.from.17.x.mdx
```diff
@@ -17,95 +17,95 @@ You can use the new `isDisabled` attribute to obtain the same behavior.
```
dpellier, line 17: would be nice to add a `## Usage changes` section first that explain how the component did move from a full-slot solution to a all-in-one component (see button or tooltip for example)

dpellier (changes requested):

packages/ods/src/components/quantity/tests/navigation/ods-quantity.e2e.ts
aesteves60: The button don't emit the focus event anymore

dpellier (approved):

Leotheluck (approved):

feoche (approved):

PR #443, opened 2024-04-29
refactor(docs): import ODS Components as Homepage elements

feoche:


dpellier (changes requested):
Links are not rendered by default on first load
Capture d’écran 2024-04-30 à 09 16 52

feoche (commented):
- [x] Title should be `40rem`
- [x] Link gap should be `2rem`

astagnol (approved):

PR #444, opened 2024-05-03
Refactor/ods toggle

aesteves60:
I add the part but not the Custom CSS example. Should Add the example or remove the possibility to custom ? 

- `hasError` props is still need ? We don't have design for that, I can add border red if it's still need


packages/ods/src/components/toggle/tests/behaviour/ods-toggle.e2e.ts
```diff
@@ -0,0 +1,110 @@
...
+    const sliderHasClassCheck = await page.evaluate(() => {
+      const slider = document.querySelector('ods-toggle')?.shadowRoot?.querySelector('.ods-toggle__container__slider');
+      return slider?.classList.contains('ods-toggle__container__slider--checked');
+    });
+    expect(await el.getProperty('value')).toBe(true);
+    expect(sliderHasClassCheck).toBe(true);
+    expect(odsChangeSpy).toHaveReceivedEventTimes(1);
+  });
+
+  it('should toggle value if disabled', async() => {
```
dpellier, line 38: `should not`

packages/ods/src/components/toggle/tests/rendering/ods-toggle.spec.ts
```diff
@@ -0,0 +1,106 @@
...
+    it('should not be set by default', async() => {
+      await setup('<ods-toggle></ods-toggle>');
+
+      expect(root?.getAttribute('value')).toBeNull();
+    });
+  });
+
+  describe('withLabel', () => {
+    it('should be reflected', async() => {
+      await setup('<ods-toggle withLabel></ods-toggle>');
```
dpellier, line 95: `with-label`

packages/storybook/stories/components/toggle/documentation.mdx
```diff
@@ -0,0 +1,36 @@
...
+
+<Canvas of={ ToggleStories.Default } sourceState='none' />
+
+<DocNavigator of={ ToggleStories } />
+
+<Markdown>{ SpecificationsToggle }</Markdown>
+
+# Examples
+
+## Usage
```
dpellier, line 23: I think the template needs to be updated, some components have this `Usage` part, some don't
Can you remove it (and change children from `###` to `##`) from the template and update all files?

packages/storybook/stories/components/toggle/migration.from.17.x.mdx
```diff
@@ -0,0 +1,87 @@
...
+## Migration examples
+
+Checked toggle:
+```html
+<osds-toggle checked>
+</osds-toggle>
+
+<!-- is now -->
+
+<ods-toggle value>
```
dpellier, line 74: this looks really weird (though it will looks better on React with value={ true })

dpellier (changes requested):

Leotheluck (changes requested):
## General
- [x] Label is not vertically aligned (the text is too close to the bottom).
- [x] Label `OFF` has a higher `z-index` than the circle.
- [ ] Animation timing seems odd (background color is updated faster than the circle transition time).
- [x] Label's text shouldn't be highlightable, use `user-select: none;`.
- [x] `isDisabled` does not change the opacity of the label `OFF`.
- [x] Design guidelines from Figma feature a box-shadow under the circle.

## Migration guide
- [x] We shouldn't put any comment such as `Design team doesn't valid this style anymore.`, rather use `This design feature has been removed in order to match the new styling of the component.`
- [x] Language mistake in `You can use the new name attribute for set an identifiant in a form.`, rather use `You can use the new 'name' attribute to identify your OdsToggle within a form.`
- [ ] Language mistake in `For add a label ON/OFF`, rather use `Will add the visual labels 'ON/OFF'`

packages/ods/src/components/toggle/src/components/ods-toggle/ods-toggle.scss
```diff
@@ -0,0 +1,94 @@
...
+        cursor: not-allowed;
+
+        & #{$slider}__label {
+          color: var(--ods-color-neutral-600);
+        }
+      }
+
+      &__label {
+        position: absolute;
+        top: 6px;
```
dpellier, line 52: Label is not more centered on FF (mac), you should not rely on px valiue to center a label as it may change between browser/os

Try something like that:
```
// This enforce the drawing zone take the whole height of the container and to start from the right sidr
position: absolute;
top: 0;
right: 0;
bottom: 0;

// Flex is the easiest way to ensure consistent vertical centering, and it works here because we works with a full height element (thx to previous block)
display: flex;
align-items: center;

// Then add a consistent spacing between the edge of the input and you label
padding-right: $toggle-label-spacing; // 4px seems to be ok
```

For ON:
```
// replace right: 0 with left: 0;
// replace padding-right: $toggle-label-spacing; with padding-left: $toggle-label-spacing;
```

packages/storybook/stories/components/toggle/toggle.stories.ts
```diff
@@ -0,0 +1,194 @@
...
+      description: 'Set a custom style properties for the slider. Example: "background-color: blue;"',
+      table: {
+        category: CONTROL_CATEGORY.design,
+        defaultValue: { summary: 'ø' },
+        type: { summary: 'string' },
+      },
+    },
+    CustomCssSliderCircleChecked: {
+      control: 'text',
+      description: 'Set a custom style properties for the slider. Example: "background-color: black;"',
```
dpellier, line 96: update each description to be more precise about which element is actually concerned by this customization

dpellier (changes requested):

Leotheluck (changes requested):
## General
- [ ] Label is not vertically aligned (the text is too close to the bottom).
- [ ] `isDisabled` does not change the opacity of the circle’s `box-shadow`.
- [ ] Circle's `box-shadow` is too strong compared to the Design on Figma.

dpellier (approved):

PR #446, opened 2024-05-10
feat(code): implement component

dpellier:
As there are no specific designs for this component, I went with the simpliest version, which is the same as the actual.
Which means no toolip on copy button (unlike clipboard), could be added on another iteration if actually asked for.

### DIFF
--- a/packages/storybook/stories/components/code/documentation.mdx
+++ b/packages/storybook/stories/components/code/documentation.mdx
@@ -1,0 +1,34 @@
+import { Canvas, Meta, Markdown } from '@storybook/blocks';
+import SpecificationsCode from '@ovhcloud/ods-components/src/components/code/documentation/spec.md?raw';
+import { Banner } from '../../banner';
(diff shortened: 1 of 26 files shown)


aesteves60 (approved):

packages/storybook/stories/components/code/documentation.mdx
```diff
@@ -0,0 +1,34 @@
...
+
+<Canvas of={ CodeStories.CustomCSS } sourceState='shown'/>
+
+# Examples
+
+## Default
+
+<Canvas of={ CodeStories.Default } sourceState='shown' />
+
+## With copy button enabled
```
Leotheluck, line 33: To match with the other documentations I would maybe only put "CanCopy", or "Can Copy"

Leotheluck (approved):

feoche (changes requested):
### Style
- [x] Padding should be 12px

feoche (approved):

PR #447, opened 2024-05-10
refactor(password): implementation component

aesteves60:


packages/ods/src/components/password/src/components/ods-password/ods-password.tsx
```diff
@@ -0,0 +1,99 @@
...
+          hasError={ this.hasError }
+          isClearable={ this.isClearable }
+          isDisabled={ this.isDisabled }
+          isLoading={ this.isLoading }
+          isMasked={ this.isMasked }
+          isReadonly={ this.isReadonly }
+          isRequired={ this.isRequired }
+          name={ this.name }
+          onOdsToggleMask={ () => this.isMasked = !this.isMasked }
+          onOdsChange={ this.onOdsChange.bind(this) }
```
dpellier, line 88: would be nice to keep the same syntax everywhere, now we usually have:
```
={ () => this.method() }
```

dpellier (changes requested):

Leotheluck (changes requested):
## General
- [ ] The Eye icon should be grey if the component is disabled. (I would assume this is also the case for the loading spinner, but this specific case is not documented on the Figma.)

## Documentation
- [ ] Instead of `You can add your own style on the password element using the part input.`, we should prefer using `You can add your own style on the password component using '::part(input)'`
- [ ] Instead of `Allow just numeric character`, we should prefer using `With this pattern setting, the component will only allow for numeric characters:`

## Migration guide
- [ ] Instead of `To add a label on the input wrapped it with the component form-field.`, we should prefer using `To add a label on a password component, wrap it within a form-field component.`
- [ ] Instead of `Form components does not provide color variant anymore. You can use custom css if you need to customize the component.`, we should prefer using `Form components do not provide color variants anymore. You can use custom CSS if you need to customise the component.`
- [ ] We should add an example of a migration from `forbidden-values` to `pattern`.

aesteves60:
* [ ]  Instead of `You can add your own style on the password element using the part input.`, we should prefer using `You can add your own style on the password component using '::part(input)'`
I let the previous version for stay the consistence with the others  component

dpellier (commented):
does the mask button needs to be disabled on readonly?
purpose of the state is to let the input content still be readable

Leotheluck (changes requested):
## Migration guide
- [ ] Form components does not provide color variant anymore. -> Form components do__ not provide color variant_s_ anymore. `(3 times)`
- [ ] Line break needed in `color` after `Has been removed`.

Leotheluck (approved):

dpellier (approved):

feoche (approved):

feoche (approved):

PR #448, opened 2024-05-14
feat(popover): implement component

dpellier:
- rename tooltip common code to overlay
- move common overlay code to utils
- update ods-button to enforce web-component height to be the same as its content


feoche (changes requested):
### Documentation
- [x] API, typo "popover can contains" → "popover can contain"
- [x] Migration, wrong label on "with-arrow", should be "new" instead of "removed"

### Demo
- [x] `position: right` is overlapped with next input
- [x] Change component implementation in demo
  - [ ] Inputs would be replaced with links
  - [ ] "Menu" Button would be labelled "Trigger" without icon

### Style
- [ ] Arrow tip does not have a shadow

### Global
- [ ] This PR changes visual UI on Clipboard demo (any browser)
Original : 
image
Modified : 
image

packages/ods/src/components/popover/tests/behaviour/ods-popover.e2e.ts
```diff
@@ -0,0 +1,52 @@
...
+    page = await newE2EPage();
+
+    await page.setContent(content);
+    await page.evaluate(() => document.body.style.setProperty('margin', '0px'));
+
+    popoverElement = await page.find('ods-popover');
+  }
+
+  beforeEach(async() => {
+    await setup(`
```
aesteves60, line 19: setup in each it for the dependence between the test

packages/ods/src/components/quantity/tests/navigation/ods-quantity.e2e.ts
```diff
@@ -44,7 +44,7 @@ describe('ods-quantity navigation', () => {
 
   describe('navigation', () => {
     it('should navigate between button & input', async() => {
-      await setup('<ods-quantity></ods-quantity>');
+      await setup('<ods-quantity></ods-quantity><button>Dummy to focus out</button>');
```
aesteves60, line 47: why ?

packages/ods/src/style/_overlay.scss
```diff
@@ -0,0 +1,19 @@
+$ods-overlay-z-index: 99;
+
+@mixin ods-tooltip() {
```
aesteves60, line 3: ods-overlay

packages/ods/src/style/_overlay.scss
```diff
@@ -0,0 +1,19 @@
...
+  z-index: $ods-overlay-z-index;
+  border-radius: 4px;
+  box-shadow: 0 2px 8px 0 rgba(0, 14, 156, 0.20);
+  background: var(--ods-color-primary-000);
+  padding: 8px;
+  color: var(--ods-color-primary-500);
+}
+
+@mixin ods-tooltip-arrow() {
```
aesteves60, line 13: ods-overlay-arrow

packages/ods/src/utils/overlay.ts
```diff
@@ -0,0 +1,124 @@
...
+    return shadowDomElement;
+  }
+
+  return hostElement;
+}
+
+function hideOverlay(popperElement: HTMLElement, cleanUpCallback?: () => void): void {
+  popperElement.style.display = 'none';
+
+  if (cleanUpCallback) {
```
aesteves60, line 53:
```suggestion
  cleanUpCallback?.()
```

packages/storybook/stories/components/popover/documentation.mdx
```diff
@@ -0,0 +1,76 @@
...
+
+<ods-input name="some-input">
+</ods-input>
+
+<ods-popover trigger-id="popover-trigger">
+  ...focusable popover content
+</ods-popover>
+```
+
+The input will get focused after the popover trigger button, causing the popover to close, thus making the element inside of it unreachable.
```
aesteves60, line 50: 👍

aesteves60 (changes requested):

packages/ods/src/components/quantity/tests/navigation/ods-quantity.e2e.ts
dpellier: Test was failing everytime locally, probably an issue when tabbing outside of the current document as there are no other target. With this we still test that the behaviour is as expected and avoid this issue

packages/ods/src/style/_overlay.scss
dpellier: not really, this style is the one for a tooltip (and a popover), but there will be other overlay who may look different, for example we may use the overlay code for the select.

packages/ods/src/style/_overlay.scss
dpellier: same idea as above

Leotheluck (approved):

feoche (approved):

aesteves60 (approved):

PR #452, opened 2024-05-17
feat(radio-button): add xs size

skhamvon:
Add 'xs' size to radio button

### DIFF
--- a/packages/themes/blue-jeans/src/index.scss
+++ b/packages/themes/blue-jeans/src/index.scss
@@ -364,6 +364,12 @@
           ),
         ),
         $radio-button: (
+          xs: (
+            border-width: calc(0.2rem * var(--ods-size-ratio)),
+            height: calc(0.75rem * var(--ods-size-ratio)),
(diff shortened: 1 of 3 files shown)


packages/themes/blue-jeans/src/index.scss
```diff
@@ -364,6 +364,12 @@
           ),
         ),
         $radio-button: (
+          xs: (
+            border-width: calc(0.2rem * var(--ods-size-ratio)),
```
astagnol, line 368: We would prefer a 2px border-with so the radio button will be 16px instead of 18px.
So, border-width: calc(0.125rem * var(--ods-size-ratio))

astagnol (changes requested):

packages/themes/blue-jeans/src/index.scss
skhamvon: fixed

astagnol (approved):

feoche (approved):

PR #455, opened 2024-05-22
Refactor/ods e2e

dpellier:
- set all suites to run with one worker
- update few focus test to prevent error on some system where tabbing out of DOM did not do anything


aesteves60 (commented):
Why not let the `max-workers=2` everywhere ?
The test will be slower

dpellier:
> Why not let the `max-workers=2` everywhere ? The test will be slower

For now to see if it fix the random failures 100% of time.

aesteves60 (approved):

Leotheluck (approved):

PR #466, opened 2024-06-12
feat(accordion): implement component

Leotheluck:
> An Accordion is a vertical header that reveals or hides an associated section of content.

### DIFF
--- a/packages/ods/src/components/accordion/src/components/ods-accordion/ods-accordion.scss
+++ b/packages/ods/src/components/accordion/src/components/ods-accordion/ods-accordion.scss
@@ -1,0 +1,79 @@
+@use '../../../../../style/focus';
+
+@keyframes toggle {
+  0% {
+    opacity: 0;
+    margin-top: -1.25rem;
+  }
+
+  100% {
+    opacity: 1;
+    margin-top: 0;
+  }
+}
+
+.ods-accordion {
+  &__wrapper {
+    transition: .15s border-color ease-out;
+    border: 1px solid;
+    border-color: transparent;
+
+    &:focus-visible,
+    &:hover {
+      border-color: var(--ods-color-primary-200);
+    }
+
+    &.ods-accordion__wrapper--is-open {
+      border-color: var(--ods-color-primary-500);
+    }
+
+    &--is-disabled {
+      opacity: .5;
+      cursor: no-drop;
+
+      &:hover,
+      &.ods-accordion__wrapper--is-open {
+        border-color: transparent;
+      }
+
+    }
+
+    &__summary {
+      display: flex;
+      position: relative;
+      align-items: center;
+      justify-content: space-between;
+      cursor: pointer;
+      padding: 24px;
+      list-style: none;
+
+      &:focus-visible {
+        @include focus.ods-focus();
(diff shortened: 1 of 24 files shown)


dpellier (changes requested):
Feels really weird to not be able to click on the summary text to toggle.
May be better to make the whole container clickable and use the same cursor (also when disabled) on the text and the surroudings.

Also the focus when opened looks really weird:
Capture d’écran 2024-06-13 à 14 11 35

packages/ods/src/components/accordion/src/components/ods-accordion/ods-accordion.scss
```diff
@@ -0,0 +1,79 @@
+@use '../../../../../style/focus';
+
+@keyframes toggle {
+  0% {
+    opacity: 0;
+    margin-top: -1.25rem;
```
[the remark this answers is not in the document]
Leotheluck: This is just to animate the accordion content so it slides from top to down, should I rather use transform translate?

astagnol (changes requested):
Capture d’écran 2024-06-17 à 14 54 19
We should not display the arrow marker of details tag (visible on Safari v17.3.1 )

packages/ods/src/components/accordion/src/components/ods-accordion/ods-accordion.tsx
```diff
@@ -0,0 +1,95 @@
...
+      <Host class='ods-accordion'>
+        <details
+          class={{
+            'ods-accordion__wrapper': true,
+            'ods-accordion__wrapper--is-disabled': this.isDisabled,
+            'ods-accordion__wrapper--is-open': this.isOpen,
+          }}
+          tabIndex={this.isDisabled ? -1 : 0}
+          onClick={(event) => this.handleClick(event)}
+          onKeyDown={(event) => this.handleKeyDown(event)}
```
dpellier, line 66: You'll have to preventDefault onKeyDown for space, to prevent page scroll (Firefox), see select or tag

packages/ods/src/components/accordion/src/components/ods-accordion/ods-accordion.tsx
```diff
@@ -0,0 +1,95 @@
...
+
+  handleKeyDown(event: KeyboardEvent): void {
+    if (event.key === ' ' || event.key === 'Enter') {
+      this.handleToggle(event);
+    }
+  }
+
+  render(): FunctionalComponent {
+    return (
+      <Host class='ods-accordion'>
```
dpellier, line 57: use `"` when direct attribute, (`'` are used in JSX interpretation, like dynamic class)

astagnol (changes requested):
- rebase with refactor/ods with new icons
- need to rework some paddings between summary and the wrapper__content, check Figma

dpellier (approved):

feoche (changes requested):
- [ ] Set icon font-size to `0.875rem`

feoche (approved):

aesteves60 (approved):

astagnol (approved):

PR #467, opened 2024-06-14
feat(card): implement component

Leotheluck:
> Card is a component designed to be a container for displaying content and other elements to communicate information or feature a call-to-action within a page.

### DIFF
--- a/packages/ods/src/components/card/src/components/ods-card/ods-card.tsx
+++ b/packages/ods/src/components/card/src/components/ods-card/ods-card.tsx
@@ -1,0 +1,22 @@
+import type { FunctionalComponent } from '@stencil/core';
+import { Component, Host, Prop, h } from '@stencil/core';
+import { ODS_CARD_COLOR, type OdsCardColor } from '../../constants/card-color';
+
+@Component({
+  shadow: true,
+  styleUrl: 'ods-card.scss',
+  tag: 'ods-card',
+})
+export class OdsCard {
+  @Prop({ reflect: true }) public color: OdsCardColor = ODS_CARD_COLOR.primary;
+
+  render(): FunctionalComponent {
+    return (
+      <Host class={ 'ods-card' }>
+        <div class={ `ods-card__wrapper ods-card__wrapper--${this.color}` } part={ 'card' }>
+          <slot></slot>
+        </div>
+      </Host>
+    );
+  }
+}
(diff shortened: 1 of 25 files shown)


packages/ods/src/components/card/src/components/ods-card/ods-card.tsx
```diff
@@ -0,0 +1,22 @@
...
+  shadow: true,
+  styleUrl: 'ods-card.scss',
+  tag: 'ods-card',
+})
+export class OdsCard {
+  @Prop({ reflect: true }) public color: OdsCardColor = ODS_CARD_COLOR.primary;
+
+  render(): FunctionalComponent {
+    return (
+      <Host class={ 'ods-card' }>
```
dpellier, line 15: direct string `class="ods-card"`

packages/ods/src/components/card/src/components/ods-card/ods-card.tsx
```diff
@@ -0,0 +1,22 @@
...
+        <div class={ `ods-card__wrapper ods-card__wrapper--${this.color}` } part={ 'card' }>
```
dpellier, line 16: why the extra div? you could just have put the style on the Host (and remove the part)

dpellier (changes requested):

packages/storybook/stories/components/card/card.stories.ts
```diff
@@ -0,0 +1,108 @@
...
+    width: 500px;
+    display: flex;
+    justify-content: center;
+    border: 3px solid green;
+  }
+</style>
+  `,
+};
+
+export const Color: StoryObj = {
```
astagnol, line 101: We should display all available colors: neutral & primary

packages/storybook/stories/components/card/card.stories.ts
```diff
@@ -0,0 +1,108 @@
...
+    },
+  }),
+  args: {
+    color: ODS_CARD_COLOR.primary,
+    content: '<ods-text preset="heading-6">Hello, world!</ods-text>',
+    customCss: '.demo-card { padding: 0 24px; }',
+  },
+};
+
+export const Default: StoryObj = {
```
astagnol, line 60: I think instead of displaying "Default" example that shows nothing, we should do something similar to Icon example (with content here)

astagnol (changes requested):

dpellier (approved):

astagnol (approved):

aesteves60 (approved):

PR #468, opened 2024-06-17
refactor/ods optimized icons

astagnol:
### Icons have been optimized as following
- they're all 16X16px
- icons as "stroke" are now vectors so, they can be scaled up or down to any size without losing quality
- icons throwing warning "Glyph bbox was incorrect" in Firefox have been fixed
- some icons have been renamed to match better with what they are displaying; it will be easier to find them
- migration guide has been updated with a comparison before/after v18 for icon names


dpellier (changes requested):
Select icons need to be updated (content value in css)

feoche (changes requested):
## Global
- [x] Use same icon in "Documentation" and "Demo"

## Demo 
- [ ] Set `alt` attribute, as it is mandatory for a11y purposes

dpellier (approved):

feoche (approved):

Leotheluck (approved):

aesteves60 (approved):

PR #469, opened 2024-06-18
feat(tabs): implement component

dpellier:


feoche (changes requested):
## Style
- [ ] If selected tab is hovered, it should stay with a blue bottom border
- [x] When in horizontal overflow state, tab labels are displayed in two lines and we can't scroll horizontally in the component itself
image

## Documentation
- [ ] Overview : First tab should be selected
- [ ] Abilities of selecting tabs on examples does not work in Chromium & Safari browsers

## Demo
- [ ] Abilities of hovering tabs does not work in Chromium & Safari browsers

packages/storybook/stories/components/tabs/tabs.stories.ts
```diff
@@ -0,0 +1,197 @@
...
+
+export default meta;
+
+export const Demo: StoryObj = {
+  render: (args) => html`
+<ods-tabs>
+  ${unsafeHTML(args.content)}
+</ods-tabs>
+  `,
+  argTypes: orderControls({
```
aesteves60, line 22: - I think it will be better to set a static content and add `isDisabled` and `isSelected` argsTypes for each tab. 
Or add an others story demo

aesteves60 (changes requested):

packages/storybook/stories/components/tabs/tabs.stories.ts
astagnol: We discussed about this yesterday with François too. We agreed that only activating one specific tab is acceptable as long as it is explained in the control description. WDYT?

feoche (approved):

aesteves60 (approved):

astagnol (approved):

Leotheluck (approved):

PR #473, opened 2024-07-01
feat(phone-number): implement component

dpellier:
**Changes on existing component:**
Select:
- added an observer to detect child modification (use case example: phone number locale change, option labels need to be updated), as sadly slot change event does not trigger on child elements updates
- styling current select is really difficult as there are no accessible parts, to manage the specific border rendering on phone-number, I've added an internal prop `borderRounded` that offers this option
- same issue with the dropdown width, when the element is large enough, it is ok to have the current behaviour where the dropdown matches the width of the select, but for the flag, we need to have a larger dropdown than the selection element. So I've added an internal prop `dropdownWidth` to manage those two cases
- exposed the custom renderer interfaces
- added `is-readonly` prop as the phone-number offers this state (=> this seems to be causing some issue with the select navigation, needs to be fixed on a separate dedicated PR)

Checkbox:
- fixed some positioning issue (the check was staying at a fixed position on a multiple select when scrolling)


astagnol (changes requested):
Capture d’écran 2024-07-02 à 10 57 38
- clearable icon button is not displayed while typing in the field. It is displayed on focus tab or when you exit the field. Same goes for Spinner when loading

Capture d’écran 2024-07-02 à 10 52 14
- we need to add a shadow (Check Flag component in Figma). We can add a box-shadow using #000D1F / ods-color-blue-900 with an opacity of 25%.

packages/ods/src/components/phone-number/src/constants/phone-number-locale.ts
```diff
@@ -0,0 +1,14 @@
+enum ODS_PHONE_NUMBER_LOCALE {
```
astagnol, line 1: 'fr' and 'en' locales only?

dpellier: these are the translated country list we have available for now

aesteves60 (changes requested):

aesteves60:
image
There are some strange think in the type generated

aesteves60 (approved):

astagnol (approved):

Leotheluck (approved):

PR #483, opened 2024-07-19
feat(doc): add search-bar migration doc

dpellier:
TBD the example uses two props from the Select that we did remove of the documentation (`border-rounded` and `dropdown-width`) as they were quite a patch due to the lack of easy style customization on select component.

But if we mention them here, we'll have to expose them in the Select documentation page also.

### DIFF
--- a/packages/storybook/stories/components/search-bar/migration.from.17.x.mdx
+++ b/packages/storybook/stories/components/search-bar/migration.from.17.x.mdx
@@ -1,0 +1,49 @@
+import { Meta } from '@storybook/blocks';
+
+<Meta title="ODS Components/Navigation/Search Bar/Migration From 17.x" />
+
+# Search Bar - migrate from v17 to v18
+----
+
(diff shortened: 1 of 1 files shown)


packages/storybook/stories/components/search-bar/migration.from.17.x.mdx
```diff
@@ -0,0 +1,49 @@
...
+
+<Meta title="ODS Components/Navigation/Search Bar/Migration From 17.x" />
+
+# Search Bar - migrate from v17 to v18
+----
+
+Search Bar has been removed from ODS components. <img src="https://img.shields.io/badge/removed-FF0000" />
+
+The component wasn't bringing any meaningful value and was not flexible enough regarding the different use cases
+that may exists.
```
Leotheluck, line 11: that may exist*

astagnol (approved):
Not a blocker but I think we should add a canvas, displaying the rendering, like we did in Menu documentation. WDYT?

aesteves60 (approved):

PR #485, opened 2024-07-23
Refactor/ods test react

aesteves60:
Make sure all form component have the same implementation for clear & reset
- event order `OdsReset` then `OdsChange`, `OdsClear` then `OdsChange`.
- If clear or reset withoutdefault value, OdsChange value should be `null`.
- After clear component sould be focus
- Radio & Checkbox add css for style focus otherwise on clear the focus is invisible

### DIFF
--- a/packages/ods/src/components/checkbox/src/components/ods-checkbox/ods-checkbox.tsx
+++ b/packages/ods/src/components/checkbox/src/components/ods-checkbox/ods-checkbox.tsx
@@ -28,10 +28,12 @@ export class OdsCheckbox {
 
   @Method()
   async clear(): Promise<void> {
+    const hasChange = this.inputEl?.checked === true;
     if (this.inputEl) {
       this.inputEl.checked = false;
     }
     this.odsClear.emit();
+    hasChange && this.onInput();
   }
 
   @Method()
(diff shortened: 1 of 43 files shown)


packages/ods/src/components/select/tests/navigation/ods-select.e2e.ts
```diff
@@ -232,7 +232,7 @@ describe('ods-select navigation', () => {
       await page.waitForChanges();
 
       expect(await isOpen()).toBe(true);
-      expect(odsChangeSpy).toHaveReceivedEventDetail({ previousValue: '', validity: {}, value: '1' });
```
dpellier, line 235: detail should always contains the same set of key, even undefine
should be 
```
expect(odsChangeSpy).toHaveReceivedEventDetail({ previousValue: undefined, validity: {}, value: '1' });
```

packages/ods/src/components/checkbox/src/components/ods-checkbox/ods-checkbox.tsx
```diff
@@ -34,6 +34,7 @@ export class OdsCheckbox {
     }
     this.odsClear.emit();
     hasChange && this.onInput();
+    this.inputEl?.focus();
```
dpellier, line 37: did you try the behaviour when using multiples checkboxes with the same name?

dpellier (changes requested):

packages/ods/src/components/checkbox/src/components/ods-checkbox/ods-checkbox.tsx
aesteves60: Yes, the focus zis set on the one where you call the clear

dpellier (approved):

PR #492, opened 2024-07-24
Refactor/ods tag

dpellier:


packages/ods/src/components/tag/src/constants/tag-size.ts
```diff
@@ -1,10 +1,9 @@
 enum ODS_TAG_SIZE {
   lg = 'lg',
   md = 'md',
-  sm = 'sm',
```
aesteves60, line 4: Not in the issue, Is it deliberate ?

dpellier: This is four different issue (one per commit), and yes the size removale is one of them

aesteves60 (approved):

PR #497, opened 2024-07-24
fix(input): change placeholder color for input and other component us…

skhamvon:
change placeholder color for input and other component using it (datepicker, password, select, textarea)


aesteves60 (approved):

astagnol (changes requested):
We should use a mixin + a token has been created on Figma for all form element placeholder color so, when those tokens have been generated, we should use it

dpellier:

PR #500, opened 2024-07-25
refactor(switch): remove props on switch item

aesteves60:


dpellier (changes requested):

dpellier:
Reapplied in https://github.com/ovh/design-system/pull/526

PR #503, opened 2024-07-25
fix(timepicker): story timepicker arg

aesteves60:
Bug in the parse of timezones who was modified by the `${arg.timezones}` like a `arg.timezones.toString()` for an array the `toString()` methods join the values with `,` so the arg was `UTC-2,UTC+2...`


dpellier (changes requested):

dpellier:
Moved to https://github.com/ovh/design-system/pull/529

PR #513, opened 2024-07-30
fix(toggle): set "off" toggle to the right position

feoche:


dpellier (approved):

packages/ods/src/components/toggle/src/components/ods-toggle/ods-toggle.scss
```diff
@@ -54,8 +54,8 @@
         right: 0;
         bottom: 0;
         align-items: center;
-        padding-right: 2px;
-        color: var(--ods-color-neutral-700);
+        padding-right: 4px;
+        color: var(--ods-color-neutral-800);
```
astagnol, line 58: I would say Design is expecting neutral-700 for default & neutral-800 while hovering or focusing

astagnol (changes requested):

astagnol (approved):

PR #536, opened 2024-08-08
refactor(components): update focus style to match hover style

dpellier:


astagnol (changes requested):
Some components are missing focus style matching hover style:
- Link
- Checkbox when checked, indeterminate
- Radio when checked

PR #548, opened 2024-08-21
fix(tabs): border on all the width

aesteves60:


packages/ods/src/components/tabs/src/components/ods-tabs/ods-tabs.scss
```diff
@@ -5,7 +5,15 @@
 
   display: inline-flex;
   flex-flow: row;
-  padding: var(--ods-outline-width);
-  width: inherit;
-  overflow-x: auto;
+  padding: var(--ods-outline-width) 0;
+  width: 100%;
+  overflow-y: auto;
```
dpellier, line 10: `overflow-x` we want an horizontal scroll (it won't impact your tab rendering with the wrapper)

packages/ods/src/components/tabs/src/components/ods-tab/ods-tab.scss
```diff
@@ -3,6 +3,7 @@
 
 :host(.ods-tab) {
   display: block;
+  margin-bottom: -2px;
```
dpellier, line 6: please use a shared variable for `margin-bottom` and `border-bottom` size here and `border-bottom` size in ods-tabs.scss

packages/ods/src/components/tabs/src/components/ods-tabs/ods-tabs.scss
```diff
@@ -5,7 +5,15 @@
...
+}
+
+.ods-tabs {
+  &-border {
```
dpellier, line 14: - use `__` for BEM block
- use a more generic name, as this block contains the tabs, it's not just for the border (something like `__items` or `__wrapper` or `__container`)

dpellier (changes requested):

dpellier (approved):

feoche (changes requested):
## Global 
- [ ] When focusing first tab, outline is being cropped on the left part

astagnol (approved):

feoche (approved):

PR #553, opened 2024-08-22
fix(select): readonly focus

aesteves60:


packages/ods/src/components/select/tests/behaviour/ods-select.e2e.ts
```diff
@@ -211,5 +211,18 @@ describe('ods-select behaviour', () => {
         });
       });
     });
+
+    describe('readonly', () => {
+      it('should emit an odsChange event', async() => {
```
dpellier, line 216: update with right description

dpellier (changes requested):
please add doc examples on select and phone number

dpellier (approved):

PR #575, opened 2024-09-27
feat(button): can display an icon on the left or right

astagnol:


packages/ods/src/components/button/src/components/ods-button/ods-button.tsx
```diff
@@ -22,6 +23,7 @@ export class OdsButton {
   @Prop({ reflect: true }) public color: OdsButtonColor = ODS_BUTTON_COLOR.primary;
   /** @docType OdsIconName */
   @Prop({ reflect: true }) public icon?: OdsIconName;
+  @Prop({ reflect: true }) public iconAlignment?: OdsButtonIconAlignment = ODS_BUTTON_ICON_ALIGNMENT.left;
```
dpellier, line 26: though it will be removed with the doc refactor, you need to add an explicit `/** @docType OdsButtonIconAlignment */` here

packages/ods/src/components/button/src/components/ods-button/ods-button.tsx
```diff
@@ -47,6 +49,7 @@ export class OdsButton {
             ods-button__button--${this.color}
             ods-button__button--${this.size}
             ods-button__button--${this.variant}
+            ${this.iconAlignment === 'right' ? 'ods-button__button--icon-right' : ''}
```
dpellier, line 52: use enum value instead of 'right'

packages/storybook/stories/components/button/button.stories.ts
```diff
@@ -64,6 +67,15 @@ export const Demo: StoryObj = {
       options: ODS_ICON_NAMES,
       description: 'See the whole list [here](/?path=/docs/ods-components-icon--documentation#name)'
     },
+    iconAlignment: {
+      table: {
+        category: CONTROL_CATEGORY.general,
+        defaultValue: { summary: ODS_BUTTON_ICON_ALIGNMENT.left },
```
dpellier, line 73: add default value to the `args` also

dpellier (changes requested):

aesteves60 (approved):

PR #583, opened 2024-10-08
feat(select): add specific style to multi-select counter

astagnol:


packages/ods/src/utils/select.ts
```diff
@@ -6,6 +6,7 @@
 function mergeSelectedItemPlugin({ label }: Record<string, unknown>): void {
   // @ts-ignore "this" is the TomSelect instance but is set as any by the lib
   const self = this; // eslint-disable-line @typescript-eslint/no-this-alias
+  const counterClassName = 'ts-merged-items__count';
```
aesteves60, line 9: Don't need to add a constant if you use it once

aesteves60 (changes requested):

dpellier (approved):

packages/ods/src/utils/select.ts
dpellier: still, other ones are defined here, so it is fine to declare all at the same place, even when used only once

aesteves60 (approved):

PR #584, opened 2024-10-08
fix(breadcrumb): set line-height on breadcrumb

astagnol:
We need to set this line-height on Breadcrumb component so, we avoid component style to be overridden by generic style (Storybook)


aesteves60 (approved):

dpellier (changes requested):
there is an issue with the link underline, very close to text and not visible at all on the icon
Capture d’écran 2024-10-09 à 09 23 43

PR #593, opened 2024-10-14
feat(button): add xs size

astagnol:
In last step regarding these tasks (add neutral color), should use new token border-radius-xs when available

### DIFF
--- a/packages/ods/src/components/button/src/components/ods-button/ods-button.tsx
+++ b/packages/ods/src/components/button/src/components/ods-button/ods-button.tsx
@@ -55,7 +55,7 @@ export class OdsButton {
             <ods-spinner
               class="ods-button__button__spinner"
               color={ ODS_SPINNER_COLOR.neutral }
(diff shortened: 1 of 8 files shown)


dpellier (changes requested):

aesteves60 (approved):

packages/ods/src/components/button/src/components/ods-button/ods-button.tsx
```diff
@@ -37,7 +37,10 @@ export class OdsButton {
   render(): FunctionalComponent {
     return (
       <Host
-        class="ods-button"
+        class={`
+          ods-button
+          ods-button__border-radius--${this.size}
```
dpellier, line 42: `ods-button--${this.size}`
(this doesn't have to be only about border-radius)

dpellier (changes requested):

dpellier (approved):

PR #596, opened 2024-10-15
fix(demo): docs form validity error on console

aesteves60:


packages/storybook/stories/components/textarea/textarea.stories.ts
```diff
@@ -18,9 +18,10 @@ export const Demo: StoryObj = {
       (() => {
```
dpellier, line 18: `async` needed

dpellier (changes requested):
Please also update control description with `Toggle this to see the component validityState`

dpellier (approved):

PR #602, opened 2024-10-22
feat(toggle): add validity state

aesteves60:


dpellier (changes requested):
New border is not replacing the previous one, resulting in different rendering and different component height, ex:
before:
Capture d’écran 2024-10-24 à 14 26 52
after:
Capture d’écran 2024-10-24 à 14 26 47

dpellier (approved):

PR #606, opened 2024-10-24
fix(select): disabled on load select

aesteves60:
When slot change, creation of the select with tom select, the props isDisabled wasn't set on the tom select


dpellier (changes requested):

dpellier (approved):

dpellier (changes requested):
waiting for style fix when select is empty

dpellier (approved):

PR #616, opened 2024-10-30
feat(global): add read-only tokens & apply style to components

astagnol:


dpellier (changes requested):
- add `isReadonly` control on select
- cursor on readonly select is a hand

packages/ods/src/components/select/src/components/ods-select/ods-select.scss
```diff
@@ -25,7 +25,7 @@ $select-padding-right: $select-caret-size + ($select-base-padding * 2);
     border: solid 1px var(--ods-color-neutral-300);
     border-radius: 0;
     background-color: $select-background-color;
-    cursor: pointer !important;
```
dpellier, line 28: still at least one use-case to handle then:
- open a select by clicking on the input => the cursor become a `text` one

dpellier (changes requested):

aesteves60 (approved):

dpellier (approved):

PR #619, opened 2024-11-04
feat(message): add non-dismissible Message

astagnol:
- also fixed an issue related to the use of ods-button as close button with its inherited height (Message content wasn't correctly aligned when a height was defined on Message itself)
- since ods-button in xs size is 24x24px, padding on the decorative icon (20px) has been kept for a better alignment

### DIFF
--- a/packages/ods/src/components/message/src/components/ods-message/ods-message.scss
+++ b/packages/ods/src/components/message/src/components/ods-message/ods-message.scss
@@ -14,7 +14,7 @@
     @include message.ods-message();
 
     &__icon {
-      padding: 0.125rem;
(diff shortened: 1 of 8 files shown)


dpellier (changes requested):
Do not keep `display: grid` with 3 expected elements when you have only 2 or the gap will be added anyway:
Capture d’écran 2024-11-04 à 14 06 08

Please add missing tests:
- new prop reflection on rendering spec suite
- describe bloc on rendering e2e suite with `should display button if ...` and `should not display button if ...`

dpellier (approved):

aesteves60 (approved):

PR #620, opened 2024-11-05
feat(modal): custom height ton part dialog

aesteves60:
You can custom the max-height of the dialog


packages/ods/src/style/_modal.scss
```diff
@@ -36,6 +36,8 @@ $ods-modal-border-radius: var(--ods-border-radius-sm);
   padding: 0;
   width: 100%;
   max-width: 512px;
+  height: 100%;
```
dpellier, line 39: this will cause issue if content is smaller

dpellier (changes requested):

dpellier (changes requested):
Careful about the height of the content, it cannot just be 100% as you have the header size to take into account
Capture d’écran 2024-11-05 à 10 39 20

astagnol (approved):

dpellier (approved):

PR #625, opened 2024-11-07
Validity state select

dpellier:


packages/storybook/stories/components/select/select.stories.ts
```diff
@@ -12,18 +12,43 @@ const meta: Meta = {
...
+        async function renderValidityState() {
+          const validity = await select.getValidity()
+          divValidityState.innerHTML = '';
+          for (let key in validity) {
+            divValidityState.innerHTML += "<div>" + key + ": " + validity[key] + "</div>";
+          }
+        }
+      })();
+    </script>`;
+    return html`,
```
aesteves60, line 36:
```suggestion
    return html`
```

aesteves60 (changes requested):
No validity state on the demo multiple ?

aesteves60 (approved):

PR #628, opened 2024-11-08
feat(text): change preset caption tag to span

astagnol:


dpellier (changes requested):
Would be nice to add somewhere in the doc that the `caption` preset is only about giving a custom style, but for actual `caption` (table) or `figcaption` (image) they should use the native tag using either an ods-text inside of it or the sass mixin

packages/storybook/stories/components/text/text.stories.ts
```diff
@@ -105,3 +106,52 @@ export const CustomCSS: StoryObj = {
...
+    <th scope="row">Chris</th>
+    <td>22</td>
+  </tr>
+  </tbody>
+</table>
+<style>
+  caption {
+    caption-side: bottom;
+  }
+  table {
```
dpellier, line 134: use a specific uniq id as this is apply to the whole page => this specifically apply on the `Properties` table on top of the page.

packages/storybook/stories/components/text/text.stories.ts
```diff
@@ -105,3 +106,52 @@ export const CustomCSS: StoryObj = {
...
+  </thead>
+  <tbody>
+  <tr>
+    <th scope="row">Chris</th>
+    <td>22</td>
+  </tr>
+  </tbody>
+</table>
+<style>
+  caption {
```
dpellier, line 131: use a specific uniq id as this is apply to the whole page

packages/storybook/stories/components/text/text.stories.ts
```diff
@@ -105,3 +106,52 @@ export const CustomCSS: StoryObj = {
...
+  tags: ['isHidden'],
+  render: () => html`
+<figure>
+  <img src="https://images.crunchbase.com/image/upload/c_pad,h_256,w_256,f_auto,q_auto:eco,dpr_1/ayzwkdawmlyzvuummuf4" alt="OVHcloud logo" />
+  <figcaption>
+    <ods-text preset="caption">My picture title</ods-text>
+  </figcaption>
+</figure>
+<style>
+  img {
```
dpellier, line 152: use a specific uniq id as this is apply to the whole page

packages/storybook/stories/components/text/text.stories.ts
```diff
@@ -32,6 +32,7 @@ export const Demo: StoryObj = {
         type: { summary: 'string' }
       },
       control: { type: 'select' },
+      description: 'Caption preset is only custom styling. For `<caption>` or `<figcaption>`, use the native tag with ods-text or Sass mixin',
```
dpellier, line 35: probably not needed here

dpellier (changes requested):

aesteves60 (approved):

PR #632, opened 2024-11-18
refactor(switch): remove ods-radio and expose methods from parent com…

dpellier:
- methods are now exposed on parent component
- odsChange events are triggered for parent with actual value and validity updated
- only one odsChange and one odsReset event is emitted on reset
- only one odsChange and one odsClear event is emitted on clear
- first item is now focused on clear


packages/ods/src/components/switch/src/components/ods-switch-item/ods-switch-item.scss
```diff
@@ -22,7 +18,13 @@
     font-weight: 600;
 
     &:not(.ods-switch-item__label--disabled) {
-      &:hover, &:focus-visible  {
+      &:focus {
```
aesteves60, line 21: We don't have this style on tabulation `:focus-visible` is required too

aesteves60 (changes requested):
When a item is set with `is-checked` we can't change the selected item

aesteves60 (changes requested):
When a item is set with `is-checked` we can't change the selected item

PR #633, opened 2024-11-18
fix(form): form element infinite loop on setAttributes & mutation observable

aesteves60:


packages/ods/src/components/datepicker/tests/behaviour/ods-datepicker.e2e.ts
```diff
@@ -217,6 +217,27 @@ describe('ods-datepicker behaviour', () => {
...
+            odsDatepicker!.value = (event.detail?.value && new Date(event.detail.value.toISOString())) ?? null;
+          }) as unknown as EventListener);
+        });
+
+        await page.evaluate((value) => {
+          document.querySelector<OdsDatepicker & HTMLElement>('ods-datepicker')!.value = new Date(value);
+        }, dummyValue);
+        await page.waitForChanges();
+
+        // expect(await el.getProperty('value')).toBe(new Date(dummyValue).toISOString());
```
dpellier, line 238: remove

packages/ods/react/tests/_app/src/components/ods-textarea.tsx
```diff
@@ -1,19 +1,39 @@
-import React from 'react-dom/client';
+import { useFormik } from 'formik';
```
dpellier, line 1: you can remove the whole Formik testing as you're able to reproduce the loop use case on native code

packages/ods/src/components/input/tests/behaviour/ods-input.e2e.ts
```diff
@@ -236,6 +236,44 @@ describe('ods-input behaviour', () => {
...
+        });
+
+        await el.setAttribute('value', dummyValue);
+        await page.waitForChanges();
+
+        expect(await el.getProperty('value')).toBe(dummyValue);
+        expect(odsValueChangeSpy).toHaveReceivedEventTimes(1);
+      });
+
+      it('should not do an infinite loop with string', async() => {
```
dpellier, line 259: `with number`

packages/ods/src/components/phone-number/tests/behaviour/ods-phone-number.e2e.ts
```diff
@@ -274,6 +274,25 @@ describe('ods-phone-number behaviour', () => {
 
         expect(odsChangeSpy).toHaveReceivedEventTimes(0);
       });
+
+      it('should not do an infinite loop', async() => {
+        const dummyValue = '0123456789';
```
dpellier, line 279: maybe best to try with the expected value instead of the unformatted one, as this may not be equal on mutation
`const dummyValue = '+33123456789';`

dpellier (changes requested):

dpellier (approved):

PR #634, opened 2024-11-19
feat(input): add search type

astagnol:


dpellier (changes requested):

aesteves60 (changes requested):
You nedd to increase the padding-right with search, clearable and is-masked
image

dpellier (changes requested):
Still KO when `isMasked` is set
Capture d’écran 2024-11-20 à 09 22 48

dpellier (approved):

aesteves60 (approved):

PR #650, opened 2024-11-29
fix(input): input type number floating number

aesteves60:


packages/storybook/stories/components/input/input.stories.ts
```diff
@@ -176,10 +176,10 @@ export const Demo: StoryObj = {
     step: {
       table: {
         category: CONTROL_CATEGORY.general,
-        defaultValue: { summary: 'ø' },
-        type: { summary: 'number' },
+        defaultValue: { summary: 'You can set the step to "any" for allowing floating value' },
```
dpellier, line 179: `Set the step to "any" to allow floating values`

dpellier (changes requested):
Setting step to `any` doesn't seems to work on Demo page, the step is using default +1
(FF mac)

PR #652, opened 2024-12-02
Has error checkables

dpellier:


packages/ods/src/components/checkbox/src/components/ods-checkbox/ods-checkbox.tsx
```diff
@@ -184,7 +185,7 @@ export class OdsCheckbox {
           aria-labelledby={ this.ariaLabelledby }
           class={{
             'ods-checkbox__checkbox': true,
-            'ods-checkbox__checkbox--error': !!this.isInvalid,
+            'ods-checkbox__checkbox--error': this.hasError || !!this.isInvalid,
```
aesteves60, line 188: Can we have the checkable component in error and checked ? Because the border red is not visible when the component is checked

dpellier: Not in a way I can think of, if those are checked, they shouldn't be in error anymore

aesteves60 (approved):

PR #667, opened 2024-12-19
fix(range): thumb clickable zone

aesteves60:


dpellier (approved):

feoche (approved):

astagnol (changes requested):
Clickable zone is now correct but I noticed this:
Capture d’écran 2024-12-20 à 11 16 17

PR #673, opened 2025-01-07
feat(modal): width height boundaries

aesteves60:
Set modal width & height according to the screen witdh


astagnol (changes requested):
I think we should add a sentence regarding the margin to the Documentation (see Placement section)

feoche (approved):

packages/storybook/stories/components/modal/documentation.mdx
```diff
@@ -79,6 +79,7 @@ They can be dismissable or not.
 When a **Modal** opening is triggered, it is displayed with a position aligned both horizontally and vertically within the viewport.
 
 **Modal** takes up 100% of the viewport width up to a maximum width of 512px.
+When the screen is a too small, then the **Modal** takes all the available space with a margin of 16px.
```
dpellier, line 82: `On small screen, the **Modal** takes all the available space with a margin of 16px.`

packages/ods/src/style/_modal.scss
```diff
@@ -48,6 +52,14 @@ $ods-modal-header-height: 32px;
     // We're not using variables because of a Chromium issue with the backdrops
     // See https://stackoverflow.com/questions/58818299/css-variables-not-working-in-dialogbackdrop/63322762#63322762
   }
+
+  @media screen and (width <= (breakpoint.$ods-breakpoint-mobile)) {
+    margin: auto 16px;
+    width: 100%;
+    min-width: fit-content;
+    max-width: fit-content;
+    max-height: revert;
```
dpellier, line 61: ?

dpellier: BTW it is often simplier to set the small CSS as default and override as the breakpoint grows

dpellier (changes requested):

dpellier (changes requested):
Missing a min-width?
Capture d’écran 2025-01-10 à 11 06 24

PR #676, opened 2025-01-19
Proposal for the meter component

BrahimS:
- Impliment meter component
- add unit and e2e test


BrahimS:
Hi @dpellier ,
Just le let you know that I've tried to rebase on master and resolve this merge conflect.
but the issues persist.

PR #681, opened 2025-01-20
feat(clipboiard): add a11y

aesteves60:


dpellier (changes requested):
- test if the clipboard value is in a formData on submit
- move clipboard to the `Form elements` doc section

dpellier (approved):

PR #687, opened 2025-01-27
Fix input clearable

aesteves60:


packages/ods/src/components/datepicker/src/components/ods-datepicker/ods-datepicker.tsx
```diff
@@ -324,8 +330,17 @@ export class OdsDatepicker implements OdsFormElement {
...
+    if (!value || isDate(new Date(value))) {
+      this.showClearable = false;
+    } else {
+      this.showClearable = true;
+    }
+  }
+
   render(): FunctionalComponent {
-    const hasClearableAction = this.isClearable && !this.isLoading && !!this.value;
+    const hasClearableAction = this.isClearable && !this.isLoading && (!!this.showClearable || !!this.value);
```
dpellier, line 343: no need to use `!!` on boolean this.showClearable

packages/ods/src/components/datepicker/src/components/ods-datepicker/ods-datepicker.tsx
```diff
@@ -324,8 +330,17 @@ export class OdsDatepicker implements OdsFormElement {
     });
   }
 
+  private onInput(event: InputEvent): void {
+    const value = (event.target as HTMLInputElement).value;
+    if (!value || isDate(new Date(value))) {
```
dpellier, line 335: Why do we hide the clearable if the date is valid?
This causes issues when for example typing `11` on a clearable datepicker, button doesn't appear

packages/ods/src/components/datepicker/tests/rendering/ods-datepicker.e2e.ts
```diff
@@ -139,4 +141,55 @@ describe('ods-datepicker rendering', () => {
...
+      await setup('<ods-datepicker is-clearable ></ods-datepicker>');
+
+      await el.type('clearable');
+      await page.waitForChanges();
+
+      buttonClearable = await page.find('ods-datepicker >>> ods-button[icon="xmark"]');
+      expect(buttonClearable).not.toBeNull();
+    });
+
+    it('should not render clearable button when value wrong format', async() => {
```
dpellier, line 189: actually it should, that's the whole point of the change, that the clearable button appear as soon as we type at least one character in the field, regarding its validity

packages/ods/src/components/input/src/components/ods-input/ods-input.tsx
```diff
@@ -78,6 +79,9 @@ export class OdsInput implements OdsFormElement {
   public async clear(): Promise<void> {
     this.odsClear.emit();
     this.value = null;
+    if (this.inputEl) {
+      this.inputEl.value = '';
```
dpellier, line 83: input element value is bind to `this.value`, we shouldn't need to update it manually after updating the props

dpellier (changes requested):

packages/ods/src/components/input/src/components/ods-input/ods-input.tsx
aesteves60: Yes, but the value is not updated for an input type number if we type a letter. So the clear doesn't work

feoche (approved):

dpellier (changes requested):
Still KO on Demo for input number
- set type to "number"
- toggle isClearable
- type "aa"
=> clearable appears
- delete "aa" (using keyboard, not the button)
=> clearable is still displayed
- type 1
=> clearable OK
- delete and type "a"
=> clearable not displayed

packages/ods/src/components/input/tests/rendering/ods-input.e2e.ts
```diff
@@ -251,6 +251,29 @@ describe('ods-input rendering', () => {
...
+      expect(buttonClearable).not.toBeNull();
+
+      await page.keyboard.press('Backspace');
+      await page.keyboard.press('Backspace');
+      await page.waitForChanges();
+
+      buttonClearable = await page.find('ods-input >>> ods-button[icon="xmark"]');
+      expect(buttonClearable).toBeNull();
+
+      // Test E2E launched with chrome. Chrome don't allow to type on input type number
```
dpellier, line 270: try typing "e" alone, it will be accepted by Chrome (as exponential symbol), but still considered an invalid value

packages/ods/src/components/input/src/components/ods-input/ods-input.tsx
```diff
@@ -258,11 +260,12 @@ export class OdsInput implements OdsFormElement {
       this.shouldUpdateIsInvalidState = false;
     }
 
-    if (!this.internals.validity?.valid && !this.internals.validity?.valueMissing) {
-      this.showClearable = true;
-    } else {
-      this.showClearable = false;
-    }
+    // console.log('this.internals.validity', this.internals.validity)
```
dpellier, line 263: remove

dpellier (changes requested):

dpellier (approved):

feoche (changes requested):
- [x] Setting "0" value in type="number" does not show isClearable button 
- [x] Setting type="number" in Demo after inputting a char in type="text" mode clears the field content, yet the isClearable button exist

astagnol (approved):

feoche (approved):

PR #709, opened 2025-02-17
refactor(ods-spinner): replaced .svg image to svg inline code

var-poro:


packages/ods/src/config/stencil.ts
```diff
@@ -5,7 +5,6 @@ import { sass } from '@stencil/sass';
 import { vueOutputTarget } from '@stencil/vue-output-target';
 import { postcss } from '@stencil-community/postcss';
 import autoprefixer from 'autoprefixer';
-import { inlineSvg } from 'stencil-inline-svg';
```
dpellier, line 8: remove the dependency in the package.json too

dpellier (changes requested):
Remove tabs commit from the PR

dpellier (approved):

astagnol (approved):

PR #714, opened 2025-02-21
feat(icon): apply a11y rules & update documentation accordingly

dpellier:


astagnol (changes requested):
Add icons (check) to the documentation (like displayed in the ticket)

var-poro (approved):

feoche (approved):

PR #715, opened 2025-02-24
feat(link): apply a11y rules & update documentation accordingly

dpellier:


feoche (changes requested):
- [ ] aria-label is not vocalized by NVDA tool. It only says "cliquable  Blank target  visité  lien" and not "Visit Example (opens in a new tab)"

var-poro (approved):

PR #717, opened 2025-02-25
A11y spinner & button

dpellier:


feoche (changes requested):
- [ ] Content written in `aria-label` is not vocalized by NVDA. Focusing Button says "cliquable  bouton" and not its related aria-label
- [ ] Spinner is vocalized "Indicateur d'occupation", but `aria-busy` is not properly vocalized in NVDA. It should vocalize "Indicateur d'occupation occupé"

PR #718, opened 2025-02-26
fix(button): adjusted height of all buttons to match Figma

var-poro:


packages/ods/src/components/button/tests/rendering/ods-button.e2e.ts
```diff
@@ -93,4 +93,85 @@ describe('ods-button rendering', () => {
...
+    it('should have the same height for all variants at size xs', async() => {
+      await setup(`
+        <div>
+          <ods-button size="${ODS_BUTTON_SIZE.xs}" variant="${ODS_BUTTON_VARIANT.default}" label="Button"></ods-button>
+          <ods-button size="${ODS_BUTTON_SIZE.xs}" variant="${ODS_BUTTON_VARIANT.ghost}" label="Button"></ods-button>
+          <ods-button size="${ODS_BUTTON_SIZE.xs}" variant="${ODS_BUTTON_VARIANT.outline}" label="Button"></ods-button>
+        </div>
+      `);
+
+      const defaultBtn = await page.find(`ods-button[size=${ODS_BUTTON_SIZE.xs}][variant="${ODS_BUTTON_VARIANT.default}"] >>> .ods-button__button`);
```
dpellier, line 107: might be better to test the `ods-button` size instead of its internal button, as we want to ensure the final rendering (the one used by integrators) has the expected height (otherwise, it may be impacted by other modifier around `button`)

packages/ods/src/components/button/tests/rendering/ods-button.e2e.ts
```diff
@@ -93,4 +93,85 @@ describe('ods-button rendering', () => {
       expect(mdButtonHeight).toBeGreaterThan(smButtonHeight);
     });
   });
+
+  describe('variants - size xs', () => {
```
dpellier, line 97: stack describe?
```
describe('variants', () => {
  describe('size xs', () => {
```

dpellier (changes requested):

dpellier (approved):

astagnol (approved):

PR #719, opened 2025-03-07
fix(combobox): refresh behavior of the list and position

var-poro:


packages/ods/src/components/combobox/src/components/ods-combobox/ods-combobox.tsx
```diff
@@ -601,24 +627,52 @@ export class OdsCombobox implements OdsFormElement {
...
       }, {
-        offset: -1, // offset the border-width size as we want it merged with the trigger.
+        offset: -1,
       });
 
-      Object.assign(this.resultListElement.style, {
-        left: '0',
-        top: `${y}px`,
-      });
+      if (!this.isOpen || !this.resultListElement) {
```
dpellier, line 652: done twice (see first instruction of RaF)

packages/ods/src/components/combobox/src/components/ods-combobox/ods-combobox.tsx
```diff
@@ -52,6 +52,7 @@ export class OdsCombobox implements OdsFormElement {
   @State() private hasNoResults: boolean = false;
   @State() private isInvalid: boolean | undefined;
   @State() private isOpen: boolean = false;
+  @State() private items: ResultElement[] = [];
```
dpellier, line 55: Looks like it is only meant to force a refresh of the element list when slot change.
Plus you just pass its ref to `this.resultElements` anyway.

If we do want to update the list without closing the list, we could just use the `forceUpdate` method to trigger a rerender.

I wonder even if it could be better to close and the reopen the dropdown when items change so that it isi visible by the user, otherwise the contents he is currently on, is suddenly modified (then the `isOpen` state will trigger the refresh)

packages/ods/src/components/combobox/src/components/ods-combobox/ods-combobox.tsx
```diff
@@ -601,24 +627,52 @@ export class OdsCombobox implements OdsFormElement {
   }
 
   private async openDropdown(): Promise<void> {
-    if (this.resultListElement && this.searchElement) {
+    if (!this.resultListElement || !this.searchElement) {
+      return;
+    }
+
+    this.resultListElement.classList.remove('ods-combobox__results--bottom', 'ods-combobox__results--top');
```
dpellier, line 634: should not be duplicated, idealy it should be done on close, and we should always close and reopen, but it's TBD as mentionned on the state comment.

packages/ods/src/components/combobox/src/components/ods-combobox/ods-combobox.tsx
```diff
@@ -601,24 +627,52 @@ export class OdsCombobox implements OdsFormElement {
...
-      this.isOpen = true;
-      this.resultListElement?.scrollTo(0, 0);
-      this.resultListElement.classList.add(`ods-combobox__results--${placement}`);
-      this.searchElement.classList.add(`ods-combobox__search--${placement}`);
-    }
+      if (placement.startsWith('top')) {
+        Object.assign(this.resultListElement.style, {
+          bottom: `${this.searchHeight}px`,
+          left: '0',
+          top: 'auto',
```
dpellier, line 660: needed?

packages/ods/src/components/combobox/src/components/ods-combobox/ods-combobox.tsx
```diff
@@ -601,24 +627,52 @@ export class OdsCombobox implements OdsFormElement {
...
+        });
+      } else {
+        Object.assign(this.resultListElement.style, {
+          bottom: 'auto',
```
dpellier, line 664: needed?

dpellier (changes requested):

packages/ods/src/components/combobox/src/components/ods-combobox/ods-combobox.tsx
```diff
@@ -366,6 +367,23 @@ export class OdsCombobox implements OdsFormElement {
     this.closeDropdown();
   }
 
+  private hasDiff(oldItems: ResultElement[], newItems: ResultElement[]): boolean {
```
dpellier, line 370: would be nice to move it into the controller and add unit tests

packages/ods/src/components/combobox/tests/rendering/ods-combobox.e2e.ts
```diff
@@ -185,6 +185,84 @@ describe('ods-combobox rendering', () => {
...
+        </div>
+      </div>
+    `);
+
+      // Focus to open dropdown
+      const inputEl = await page.find('ods-combobox >>> ods-input');
+      await inputEl.click();
+      await page.waitForChanges();
+
+      // Verify dropdown is open
```
dpellier, line 210: keep only meaningful comments

packages/ods/src/components/combobox/tests/rendering/ods-combobox.e2e.ts
```diff
@@ -185,6 +185,84 @@ describe('ods-combobox rendering', () => {
...
+        const searchEl = document.querySelector('ods-combobox')?.shadowRoot?.querySelector('.ods-combobox__search');
+        return results?.classList.contains('ods-combobox__results--top') &&
+             searchEl?.classList.contains('ods-combobox__search--top');
+      });
+
+      expect(isPositionedTop).toBe(true);
+    });
+
+    it('should maintain correct positioning when reopening the dropdown', async() => {
+    // Similar setup as above
```
dpellier, line 225: picture each tests as independant, if test above is removed one day, this comment gets wrong

packages/ods/src/components/combobox/tests/rendering/ods-combobox.e2e.ts
```diff
@@ -228,4 +306,4 @@ describe('ods-combobox rendering', () => {
       expect(hasScroll).toBe(false);
     });
   });
-});
+});
```
dpellier, line 309: check your IDE to avoid removing last line automatically (or we'll have a diff each time)

dpellier (changes requested):

dpellier (approved):

PR #730, opened 2025-03-26
fix(clipboard): reset padding when isMasked isnt defined

var-poro:


feoche (changes requested):
- [ ] Overflowing content is displayed when isMasked is set to true

dpellier (approved):

feoche (approved):

astagnol (changes requested):
There is still an issue in my opinion.

See the Clipboard with its copy button only:
Screenshot 2025-03-26 at 11 03 32

Now, here is the result with both buttons:
Screenshot 2025-03-26 at 11 03 44

We should have a padding between Clipboard content and the first displayed button.

var-poro:
Problem comes from the input component, same here with password: 
Screenshot 2025-03-26 at 13 49 43

astagnol (approved):

PR #731, opened 2025-03-26
Ods react doc generator

dpellier:
Doc generation is using two generators:
- embedded storybook react-docgen that already provides all needed info to display components props table
- typedoc to generate information about non-component files (for example, the enum files)

Latest version of storybook is 8.6.9, though the following issue https://github.com/storybookjs/storybook/issues/30919 prevents us to use it, so we stick to latest stable for us (8.6.7).

Though this version does come with a weird random issue on Technical Information rendering, which may causes the example canvas to be rendered the wrong way for a few seconds (see https://github.com/storybookjs/storybook/issues/30356). If this is not fixed before our final release, we may need to downgrade even more.

### DIFF
--- a/package.json
+++ b/package.json
@@ -14,10 +14,10 @@
   ],
   "scripts": {
     "build:ci": "lerna run --stream build:ci",
-    "build:prod": "lerna run --stream build:prod && yarn run doc",
+    "build:prod": "lerna run --stream build:prod",
     "build:storybook": "lerna run --stream build:storybook",
     "clean": "lerna run clean",
-    "doc": "lerna run doc && lerna run copy:doc",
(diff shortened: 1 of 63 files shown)


var-poro:
https://github.com/storybookjs/storybook/pull/30929

Issue has been apparently patched, so we could try getting on the latest version

dpellier:
> [storybookjs/storybook#30929](https://github.com/storybookjs/storybook/pull/30929)
> 
> Issue has been apparently patched, so we could try getting on the latest version

👍  Just have to wait a bit as 8.9.10 is not released yet

var-poro (approved):

PR #735, opened 2025-03-28
Ods React CI workflows

dpellier:


.cds/actions/setup.yml
var-poro, line 1: Is --force really useful here?

dpellier: We used to have some issue on previous pipeline on that specific step beofre we enforce the focred fetch

var-poro (approved):

PR #737, opened 2025-04-01
Ods react overlays

dpellier:


var-poro (approved):

feoche:
NVDA does not vocalize Popover nor Tooltip content due to the fact that it uses `aria-describedby` instead of `aria-labelledby`

feoche:
NVDA does not vocalize Popover nor Tooltip content due to the fact that it uses `aria-describedby` instead of `aria-labelledby`

PR #744, opened 2025-04-07
fix(combobox): ensure 'add new entry' option visibility when no items…

var-poro:
… are present


dpellier (changes requested):
Got an issue, when doing the following:
- empty combobox
- type a => Add appear
- delete a => Add stays
Screenshot 2025-04-07 at 17 04 29

astagnol (approved):

dpellier (approved):

PR #746, opened 2025-04-07
feat(input): implement component

dpellier:


astagnol (changes requested):
- search icon button & masked button should be in primary color
- maybe we should add some space between the inputs for Types Story ?
- I would have add a value for all form element components in readonly. WDYT ?

var-poro (approved):

PR #766, opened 2025-04-23
feat(radio): implement component

dpellier:


packages/ods-react/src/components/radio-group/src/components/radio-label/RadioLabel.tsx
```diff
@@ -0,0 +1,26 @@
+import { RadioGroup } from '@ark-ui/react/radio-group';
+import { type ComponentPropsWithRef, type FC, type JSX, forwardRef } from 'react';
+
+interface RadioLabelProp extends ComponentPropsWithRef<'span'> {}
```
var-poro, line 4: disabled state should also apply on label, so `cursor: not-allowed` displays when hovering the disabled radio's label.

packages/ods-react/src/components/radio-group/src/dev.module.css
```diff
@@ -0,0 +1,11 @@
+.custom-radio {
+  gap: 16px;
+}
+
+.custom-radio-group {
+  background-color: #4bb2f6;
```
var-poro, line 6: we might want to use var(--ods-color-primary-300) to support theming

var-poro (changes requested):

astagnol (changes requested):
Radio full size with its border should be 16X16px

packages/ods-react/src/components/radio-group/src/dev.module.css
dpellier: This is the dev file, only used for our own internal development, we don't export this

var-poro (approved):

astagnol (approved):

PR #767, opened 2025-04-23
feat(message): implement component

dpellier:


astagnol (approved):

packages/storybook/stories/components/message/message.stories.tsx
```diff
@@ -0,0 +1,113 @@
...
+      options: MESSAGE_VARIANTS,
+    },
+  }),
+  args: {
+    children: 'My message',
+  },
+};
+
+export const Color: Story = {
+  decorators: [(story) => <div style={{ display: 'flex', flexFlow: 'row', gap: '8px', alignItems: 'center' }}>{ story() }</div>],
```
var-poro, line 52: We used to have flex-direction: column which was cleaner in presentation imo

var-poro (approved):

packages/storybook/stories/components/message/message.stories.tsx
dpellier: updated

PR #784, opened 2025-05-14
feat(file-upload): implement component

dpellier:


astagnol (approved):

packages/ods-react/src/components/file-upload/src/constants/file-error.ts
```diff
@@ -0,0 +1,17 @@
+enum FILE_REJECTION_CAUSE {
+  invalidFile = 'invalid-file',
+  maxFileReached = 'max-file-reached',
+  sizeTooLarge = 'size-too-large',
+  unknownError = 'unknown-error',
+  wrongFormat = 'wrong-format',
```
var-poro, line 6: is there an ARK error available when dropping a directory? It feels weird to have no feedback as a user.

var-poro (changes requested):

var-poro (approved):

PR #791, opened 2025-05-22
feat(timepicker): implement component

dpellier:


var-poro (changes requested):
not related to timepicker but select : chevron is 2px bigger in the newer version, is this intended?

packages/ods-react/src/components/form-field/src/contexts/withFormField.tsx
```diff
@@ -0,0 +1,31 @@
...
+  WrappedComponent: ForwardRefExoticComponent<PropsWithoutRef<P> & RefAttributes<T>>,
+): ForwardRefExoticComponent<PropsWithoutRef<P> & RefAttributes<T>> {
+  function WithFormFieldComponent({ id, invalid, ...props }: P, ref: ForwardedRef<T>): JSX.Element {
+    const fieldContext = useFormField();
+
+    return (
+      <WrappedComponent
+        { ...(props as PropsWithoutRef<P>) }
+        id={ id || fieldContext?.id }
+        invalid={ invalid || fieldContext?.invalid }
```
var-poro, line 19: This may result in overwritting values, ie `invalid === false`. 
We may want to make it more robust by using `invalid={ invalid !== undefined ? invalid : fieldContext?.invalid }`

PR #793, opened 2025-05-22
feat(switch): implement component

dpellier:


packages/storybook/stories/components/switch/switch.stories.tsx
```diff
@@ -0,0 +1,126 @@
...
+import { CONTROL_CATEGORY } from '../../../src/constants/controls';
+import { excludeFromDemoControls, orderControls } from '../../../src/helpers/controls';
+
+type Story = StoryObj<SwitchProp>;
+
+const meta: Meta<SwitchProp> = {
+  argTypes: excludeFromDemoControls(['defaultValue', 'onValueChange', 'value']),
+  component: Switch,
+  subcomponents: { SwitchItem },
+  title: 'ODS Components/Switch',
```
var-poro, line 13: ODS Components/Form Elements/Switch

packages/ods-react/src/components/switch/src/components/switch/switch.module.scss
```diff
@@ -0,0 +1,26 @@
...
+
+@layer ods-atoms {
+  .switch {
+    box-sizing: border-box;
+    display: inline-flex;
+    gap: 4px;
+    align-items: center;
+    justify-content: center;
+    border-radius: var(--ods-border-radius-lg);
+    background-color: var(--ods-color-neutral-100);
```
var-poro, line 12: Add font-family: inherit; to avoid button's resets

packages/ods-react/src/components/switch/src/components/switch/Switch.tsx
```diff
@@ -0,0 +1,64 @@
...
+
+interface SwitchValueChangeDetail {
+  value: string,
+}
+
+interface SwitchProp extends ComponentPropsWithRef<'div'> {
+  defaultValue?: string,
+  disabled?: boolean,
+  invalid?: boolean,
+  onValueChange?: (detail: SwitchValueChangeDetail) => void,
```
var-poro, line 15: Since we use default "value" prop, shouldn't we use onChange instead of onValueChange?

var-poro (changes requested):

packages/storybook/stories/components/switch/switch.stories.tsx
dpellier: it's not a form element anymore, those are just buttons
it was using an input on previous version, but it was always weird

packages/ods-react/src/components/switch/src/components/switch-item/switchItem.module.scss
```diff
@@ -0,0 +1,41 @@
...
+      @include focus.ods-focus();
+
+      background-color:  var(--ods-color-primary-100);
+    }
+
+    &[data-state="on"] {
+      background-color:  var(--ods-color-primary-600);
+      color:  var(--ods-color-primary-000);
+
+      &[data-disabled] {
```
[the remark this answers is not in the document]
dpellier: We'll go through every component regarding this when the final combination rule bteween disabled/invalid/readonly will be updated.
Invalid is removed anyway from this component as it is not a form element anymore.

packages/ods-react/src/components/switch/src/components/switch/Switch.tsx
var-poro: We do not want to lose the [native onChange](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Global_attributes) so let's keep it that way

var-poro (approved):

astagnol (approved):

PR #799, opened 2025-06-03
feat(icon): add updated icons with correct ratio

astagnol:


packages/ods-react/src/components/tag/src/components/tag/tag.module.scss
```diff
@@ -42,14 +42,6 @@
       }
     }
 
-    &--lg {
-      @include tag.ods-tag-size('lg');
```
dpellier, line 46: sure about that? this also remove 
```
padding: 2px 10px;
height: 28px;
```
which is quite a desogn breaking change regarding lg tag

dpellier (approved):

var-poro (approved):

PR #816, opened 2025-06-20
feat(combobox): implement multiple

var-poro:


dpellier (changes requested):

dpellier (commented):
- Combobox whole element height should be `32px`, not `38`.

When in readonly mode:
- tags should not be clickable
- keyboard nav should be disabled?? => TBD

Issue with tag focus :
- go to Demo
- set multiple
- pick three value
- move using keyboard to the first value
- click on the second tag to remove it
- press enter
=> nothing happens, though the first tag still appears selected

astagnol (changes requested):
With Voiceover, Combobox is recognized as a "list box pop-up, **button**", not a "list box pop-up, **combo box**" as it is in the original lib. It causes accessibility issues since when the user is typing, nothing is vocalized and the same goes for the list of options.

dpellier (approved):

PR #833, opened 2025-07-03
fix(skeleton): update the background value so it respects the skeleton height

StephaneRavet:
### Pull Request Description

This PR improves the `<OdsSkeleton />` component to ensure that its background correctly fills the entire height of the component, even when a custom height is applied via a CSS class or inline style.

- The component now visually adapts to any height set by the user.
- This guarantees consistent appearance when using utility classes or custom styles.

#### Example usage

```jsx
<OdsSkeleton className="h-[40px]" />
```

In this example, the Skeleton component will have a height of 40px, and its background will fully fill the component.


dpellier:
Seen together, this issue is related to v18.
We'll add it to the v18 backlog.
In the meantime, you can override the skeleton internal style using the `::part(skeleton)` css pseudo-selector to fix your local issue.

PR #837, opened 2025-07-04
feat(skeleton): improve accessibility & add examples in documentation

astagnol:


packages/ods-react/src/components/skeleton/src/components/skeleton/Skeleton.tsx
```diff
@@ -14,7 +14,8 @@ const Skeleton: FC<SkeletonProp> = forwardRef(({
       className,
     )}
     ref={ref}
-    {...props}>
+    {...props}
+    aria-hidden="true">
```
dpellier, line 18: even if `aria-hidden` should always be true, it may be better to set it above the props extend (people may want to be creative for some reason with their usage of skeleton)

packages/storybook/stories/components/skeleton/skeleton.stories.tsx
```diff
@@ -26,3 +26,13 @@ export const Overview: Story = {
     <Skeleton />
   ),
 };
+export const AccessibilityLoading: Story = {
+  tags: ['!dev'],
+  render: ({}) => (
+    <>
```
dpellier, line 32: remove `<></>`

dpellier (changes requested):

var-poro (approved):

PR #852, opened 2025-07-11
feat(pagination): add a11y changes and documentation

dpellier:


var-poro (approved):

astagnol (changes requested):
How about "    On Pagination's `select`, add `aria-labelledby` containing the `select` id + `before-total-items` + `after-total-items`

It helps users understand their position within the paginated content, by vocalizing "10 of 100 results" for example" ?

PR #860, opened 2025-07-17
fix(code): add example with restricted height and update styles

var-poro:


dpellier (changes requested):
Issue on single line
18.6.3:
Screenshot 2025-07-18 at 09 52 22

Your branch:
Screenshot 2025-07-18 at 09 52 27

PR #866, opened 2025-07-19
Release/18.6.4

Poppyto:
Resolve this bug :

If you add a file in file upload, then cancel it, then re-add the same file, it will not re-add it


Poppyto:
Ok thank you, nice solution :)

PR #870, opened 2025-07-21
feat(ods-react): bump ark to 5.17.0

dpellier:
What may concern us (thus need to be check for non-regression):
```
General:
- Ensure pointerdown or click event handlers only execute when the main button is clicked

Combobox:
- Expose reason to onOpenChange and onInputValueChange callbacks
- Expose api.clearHighlightedValue function to clear highlighted value
- Fix issue where rehydrating defaultValue or value after fetching items doesn't update the valueAsString
- Fixed issue where onInputValueChange could be called twice when selecting an item
- Fixed issue where combobox with allowCustomValue: true used within in a form requires two enter keypress to submit


Date Picker
- Fix issue where datepicker errors when setting selectionMode=range and minView=year
- Fix issue where focusedValue could not be fully controlled
- Add support for inline prop to render the date calendar inline
- Fix issue with keyboard selection where setting unavailable date causes month view to behave differently from clicking with mouse

File Upload:
- Export FileError, FileMimeType, and FileRejection types
- Fix issue where calling api.setFiles invokes validation with incorrect acceptedFiles
- Fix issue where the browser might not be able to infer the mime type of a file due to limitations, drag source or security restrictions. As a fallback in the file validation logic, we now infer the mime type from the file extension.
- Prevented undefined in acceptedFiles when no files accepted

Number Input:
- Fix issue where default pattern does not allow negative numbers with decimal point

Popover:
- Fixed issue where onOpenChange could be called twice when controlled

Select:
- Select highlighted item only if it exists in the collection
- Expose api.clearHighlightedValue function to clear highlighted value
- Fix issue where rehydrating defaultValue or value after fetching items doesn't update the valueAsString
- Fixed issue where highlighted item could be cleared when navigating up/down the list with keyboard

Tabs:
- Fixed issue where tabs with links should not trigger tab change upon cmd/middle click
```


var-poro (commented):
Bump to 5.18?
`Date Picker: Fix date comparison issues when time components are involved. This resolves critical issues with date
comparison operations when different date types (CalendarDate, CalendarDateTime, ZonedDateTime) are mixed,
particularly in scenarios involving time components.`

PR #871, opened 2025-07-21
feat(tokens): add new design tokens and update message bg-color for b…

astagnol:
…etter contrast


packages/ods-react/src/components/modal/src/components/modal-header/modalHeader.module.scss
```diff
@@ -18,6 +18,12 @@
 
     &--primary {
       @include modal.ods-modal-color('primary');
+
+      button {
```
dpellier, line 22: no generic selector, add a dedicated class `__close`

dpellier (changes requested):

var-poro (approved):

packages/ods-react/src/components/modal/src/components/modal-header/modalHeader.module.scss
```diff
@@ -19,7 +19,7 @@
     &--primary {
       @include modal.ods-modal-color('primary');
 
-      button {
+      .modal__close {
```
dpellier, line 22: `&__close`
(actually here you will have to do like in message, with a var reference on top `#{$message}__close`)

packages/ods-react/src/components/modal/src/components/modal-header/ModalHeader.tsx
```diff
@@ -31,6 +31,7 @@ const ModalHeader: FC<ModalHeaderProp> = ({
         <Dialog.CloseTrigger asChild>
           <Button
             aria-label={ translate(MODAL_I18N.closeButton) }
+            className={ style['modal__close'] }
```
dpellier, line 34: `style['modal-header__close']`

packages/ods-react/src/components/modal/src/components/modal-header/modalHeader.module.scss
dpellier: this is why you descend from parent name

dpellier (changes requested):

dpellier (approved):

PR #885, opened 2025-07-28
Fix file upload label

var-poro:


dpellier (changes requested):

packages/storybook/stories/components/file-upload/file-upload.stories.tsx
```diff
@@ -1,3 +1,4 @@
+import { FormField, FormFieldLabel } from '@ovhcloud/ods-react-form-field/src';
```
dpellier, line 1: Prefer to use `from '../../../../ods-react/src/components/form-field/src';` when using actually using code in a Story to avoid storybook to use the minified version, which can cause some error / minification rendering

dpellier (changes requested):

dpellier (approved):

feoche (approved):

PR #900, opened 2025-08-01
feat(datepicker): add new format prop and usage examples

var-poro:


dpellier:
Won't do for now, as modifying `format` won't solve the typing issue

PR #914, opened 2025-08-13
feat(doc): add live editor on Demo page

dpellier:


astagnol (changes requested):
Updating any control (boolean, input...) for form elements reset the component (typed value, checked, unchecked etc.)

var-poro (approved):

dpellier:
Closed for now as the editor plugin does re-render on every change, so there is no easy way to prevent component manual change to disappear when updating controls or the source.

PR #924, opened 2025-09-09
feat(theme): expose fonts separately from design tokens

dpellier:
Based on https://github.com/ovh/design-system/pull/920.

- theme style and fonts are separated.
- fonts are imported using woff files.
- font files are exposed.
- default stays the same as the actual import.
- we're using the variant font file instead of multiple weight files
- the `source-sans-pro` is actually a custom one, with a different line-height.
- new `normalize-css` import available from `@ovhcloud/ods-react` package (instead of the documentation section about font-family inheritance for native element)
- removed the Guide section as:
  - the font list is actually just variant files
  - preloading font is a bundle optimization relative to each projects build (and may not be always advised as it may take away browser resources from the loading of other resources)


Ennoriel (commented):
Great PR ;)
Though, it would be more rewarding for the author of the PR to keep the initial commit (PR https://github.com/ovh/design-system/pull/920)

PR #928, opened 2025-09-25
Popover new props

dpellier:


feoche (changes requested):
- [x] Autofocus seem to be triggered at all times, even if set to false
- [ ] Documentation : missing examples of fullWidth + autofocus

astagnol (approved):

PR #933, opened 2025-09-30
feat(doc): format code before display in code sandbox

dpellier:


var-poro (changes requested):
Not related to this particular PR but default Accordion story crashes, missing import Accordion

astagnol (approved):

PR #1051, opened 2026-03-25
feat(markdown): implement component

dpellier:
This PR includes:
- new component Markdown
- refactor of Text typing to allow better generic management
- new preset blockquote added to Text

### DIFF
--- a/packages/ods-react/src/components/markdown/src/dev.module.css
+++ b/packages/ods-react/src/components/markdown/src/dev.module.css
@@ -1,0 +1,3 @@
+.custom-markdown {
+
+}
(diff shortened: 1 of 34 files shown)


astagnol (approved):

packages/ods-react/src/components/markdown/src/dev.module.css
```diff
@@ -0,0 +1,3 @@
+.custom-markdown {
+
+}
```
crotru, line 3: Maybe delete this file

crotru (approved):

packages/ods-react/src/components/markdown/src/dev.module.css
dpellier: We tend to keep them to make it more easier for future dev to add CSS customization dev test without having to recreate the file

PR #1060, opened 2026-04-08
feat(pagination): refactor to compound element && add page selector feature

dpellier:
Pagination usage as a single component is now deprecated.
Correct usage now uses compound components through:
- PaginationPages
- PaginationPageSelector
- PaginationPageSizeSelector
Previous behavior still works as expected though.

### DIFF
--- a/packages/ods-react/src/components/pagination/src/components/pagination-page-selector/PaginationPageSelector.tsx
+++ b/packages/ods-react/src/components/pagination/src/components/pagination-page-selector/PaginationPageSelector.tsx
@@ -31,10 +31,10 @@ const PaginationPageSelector: FC<PaginationPageSelectorProp> = forwardRef(({
     event.preventDefault();
     event.stopPropagation();
     const formData = new FormData(event.target as HTMLFormElement);
-    const newPage = formData.get(INPUT_NAME) as number | null;
+    const newPage = Number(formData.get(INPUT_NAME));
 
-    if (newPage && newPage !== page) {
-      setPage(Math.min(newPage, totalPages));
+    if (!isNaN(newPage) && newPage !== page) {
+      setPage(Math.max(Math.min(newPage, totalPages), 1));
     }
   }
 
--- a/packages/ods-react/src/components/pagination/src/components/pagination-page-selector/PaginationPageSelector.tsx
+++ b/packages/ods-react/src/components/pagination/src/components/pagination-page-selector/PaginationPageSelector.tsx
@@ -1,0 +1,72 @@
+import { usePaginationContext } from '@ark-ui/react/pagination';
+import classNames from 'classnames';
+import { type ComponentPropsWithRef, type FC, type FormEvent, type JSX, type ReactNode, forwardRef, useId } from 'react';
+import { BUTTON_SIZE, Button } from '../../../../button/src';
+import { INPUT_TYPE, Input } from '../../../../input/src';
+import style from './paginationPageSelector.module.scss';
+
+interface PaginationPageSelectorProp extends ComponentPropsWithRef<'form'> {
+  /**
+   * The label displayed near the go-to-page input.
+   */
+  label?: ReactNode;
+  /**
+   * The label displayed in the go-to-page submit button.
+   */
+  submitLabel?: ReactNode;
+}
+
+const INPUT_NAME = 'page';
+
+const PaginationPageSelector: FC<PaginationPageSelectorProp> = forwardRef(({
+  className,
+  label = 'Go to page',
+  submitLabel = 'Go',
+  ...props
+}, ref): JSX.Element => {
+  const { page, setPage, totalPages } = usePaginationContext();
+  const textId = useId();
+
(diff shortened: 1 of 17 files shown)


astagnol (approved):

packages/ods-react/src/components/pagination/src/components/pagination-page-selector/PaginationPageSelector.tsx
```diff
@@ -0,0 +1,72 @@
...
+  const textId = useId();
+
+  function handleSubmit(event: FormEvent): void {
+    event.preventDefault();
+    event.stopPropagation();
+    const formData = new FormData(event.target as HTMLFormElement);
+    const newPage = formData.get(INPUT_NAME) as number | null;
+
+    if (newPage && newPage !== page) {
+      setPage(Math.min(newPage, totalPages));
```
crotru: Maybe `Number()` rather than type casting, so Math.min get an actual number.

packages/ods-react/src/components/pagination/src/components/pagination-page-selector/PaginationPageSelector.tsx
```diff
@@ -0,0 +1,72 @@
...
+        id={ textId }>
+        { label }
+      </span>
+
+      <Input
+        aria-labelledby={ textId }
+        className={ style['pagination-page-selector__input'] }
+        name={ INPUT_NAME }
+        type={ INPUT_TYPE.number } />
+
```
crotru, line 59: Shall this hold a `min` and `max` attributes ?
(not sure about this one...)

packages/ods-react/src/components/pagination/src/components/pagination-page-selector/paginationPageSelector.module.scss
```diff
@@ -0,0 +1,16 @@
+@layer molecules {
```
crotru: `@layer ods-molecules {`

packages/ods-react/src/components/pagination/src/components/pagination-pages/paginationPages.module.scss
```diff
@@ -0,0 +1,9 @@
+@layer ods-molecules {
+  .pagination-pages {
+    display: flex;
+    flex-direction: row;
+    align-items: center;
+    margin: 0;
```
crotru: Might be constraining

packages/ods-react/src/components/pagination/src/components/pagination-pages/paginationPages.module.scss
```diff
@@ -0,0 +1,9 @@
...
+    font-family: inherit;
```
crotru: Implicit

packages/ods-react/src/components/pagination/src/components/pagination-pages/PaginationPages.tsx
```diff
@@ -0,0 +1,70 @@
...
+              <PaginationItem key={ index } page={ page } index={ index } disabled={ disabled } />
+            ) : (
+              <VendorPagination.Ellipsis
+                asChild
+                index={ index }
+                key={ index }>
+                <Button
+                  disabled
+                  variant={ BUTTON_VARIANT.ghost }>
+                  &#8230;
```
crotru, line 46: Why not literal `…`?

packages/ods-react/src/components/pagination/src/components/pagination-pages/paginationPages.module.scss
crotru: (font-family in pagination.module.css can be removed too)

crotru (changes requested):

packages/ods-react/src/components/pagination/src/components/pagination-page-selector/PaginationPageSelector.tsx
dpellier: Personaly I prefer to handle the out-of-bounds value using fallback to actual limits (like typing 555 go to page 50, typing -1 go to page 1).
If we add explicit min/max, it will prevent the submit action until user fix its value.
But this can be discussed further.

packages/ods-react/src/components/pagination/src/components/pagination-pages/PaginationPages.tsx
dpellier: So it is managed as a single character (I guess, not really sure why it was setup like that TBH), but this does render a bit differently (tested in FF), so I think we'll keep it like this for now.

packages/ods-react/src/components/pagination/src/components/pagination-page-selector/PaginationPageSelector.tsx
crotru: Ok with that (plus this would imply an error state, an error text, ...)

crotru (approved):

PR #1062, opened 2026-04-09
fix(accordion): allow trigger text to be selected

dpellier:
Behavior is a bit different on Safari as it allows selection of button content only if the selection drag starts from an actual text element (for example, it you select the accordion content and the trigger)

### DIFF
--- a/packages/ods-react/src/components/accordion/src/components/accordion-trigger/accordionTrigger.module.scss
+++ b/packages/ods-react/src/components/accordion/src/components/accordion-trigger/accordionTrigger.module.scss
@@ -15,6 +15,7 @@
     color: inherit;
     font-family: inherit;
     font-size: inherit;
+    user-select: text;
 
     &:disabled {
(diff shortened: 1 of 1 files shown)


packages/ods-react/src/components/accordion/src/components/accordion-trigger/accordionTrigger.module.scss
```diff
@@ -15,6 +15,8 @@
     color: inherit;
     font-family: inherit;
     font-size: inherit;
+    -webkit-user-select: text; // Needed for Safari
```
crotru: I don't see the difference with prod on safari.
Not sure it is needed.
The behavior of _being able to select interactive dom content when start dragging from a non-interactive element_ is already there on safari.

crotru (changes requested):

crotru (approved):

astagnol (approved):

PR #1063, opened 2026-04-09
fix(cart): align item price to the right

dpellier:


crotru (commented):
That fits the price at the right place, but should the remove button vertically align with it ?

crotru (approved):

astagnol (approved):

PR #1080, opened 2026-04-29
Drawer backdrop & recipe script fixes

dpellier:


crotru (approved):

feoche (approved):

var-poro (approved):

packages/ods-react/src/components/drawer/src/components/drawer-content/DrawerContent.tsx
```diff
@@ -25,11 +25,24 @@ const DrawerContent: FC<DrawerContentProp> = forwardRef(({
...
+        <Dialog.Backdrop
+          className={ style['drawer-backdrop'] }
+          style={{
+            zIndex: 'calc(var(--ods-theme-overlay-z-index) + 1)',
+            ...(backdropStyle || {}),
+          }} />
+      }
+
+      <Dialog.Positioner style={{
+        zIndex: 'calc(var(--ods-theme-overlay-z-index) + 2)',
```
var-poro: This inline z-index turns the positioner into a stacking context, which silently neutralizes `--ods-drawer-z-index` — a variable still applied by the `ods-drawer()` mixin (`_drawer.scss:18`) and still published in the component's `cssVariable.json` (so it appears in the public Technical Information table). Overriding it no longer moves the drawer relative to the rest of the page, only within its own positioner.

Modal went through the same migration but did it fully: `$ods-modal-z-index` carries an explicit `// [Deprecated]` comment and no z-index variable is exposed in Modal's public `cssVariable.json`. Suggest doing the same here (deprecation marker + doc update), and applying the positioner z-index **only when `backdrop` is set** — otherwise every existing drawer without backdrop changes stacking behavior, which we want to avoid.

var-poro: Related: the `_overlay.scss` scale explicitly places the drawer **above** the modal (`$ods-overlay-drawer-z-index = $ods-overlay-modal-z-index + 1`), but this `overlay + 2` is exactly Modal's positioner value — so a drawer and a modal now tie, and DOM order decides who wins. Since Modal already deprecated its slot in that scale, the drawer slot should be deprecated too rather than silently contradicted.

packages/storybook/stories/components/drawer/drawer.stories.tsx
```diff
@@ -109,6 +109,28 @@ export const AnatomyTech: Story = {
   ),
 };
 
+export const Backdrop: Story = {
```
var-poro, line 57: The drawer has a `tests/rendering` e2e suite — a cheap test asserting the backdrop is rendered when the prop is set (and absent otherwise) would match the repo's conventions.

packages/storybook/stories/components/drawer/technical-information.mdx
```diff
@@ -56,6 +56,10 @@ import * as DrawerStories from './drawer.stories';
 
 <Canvas of={ DrawerStories.Default } sourceState="shown" />
 
+<Heading label="Backdrop" level={ 3 } />
```
var-poro: Coordination with #1097: that PR removes the Examples section from `technical-information.mdx` in favor of a dedicated `examples.mdx` per component. Whichever merges second will need to move this block.

var-poro (commented):
The backdrop implementation is a faithful copy of Modal's mechanics (same inline `overlay+1`/`overlay+2` scheme, same `backdropStyle` escape hatch) — good consistency. The recipe script fixes are correct too: `dirent.path` is deprecated and the repo requires Node ≥ 22, so `parentPath` is safe.

The main ask (see inline comments) is to finish the migration the way Modal did it: Modal deprecated `$ods-modal-z-index` with an explicit comment and removed the z-index variable from its public `cssVariable.json`, while here the documented `--ods-drawer-z-index` gets silently neutralized — and the new positioner z-index also applies to existing drawers without backdrop, which changes their stacking behavior.

packages/ods-react/src/components/drawer/src/components/drawer-content/DrawerContent.tsx
var-poro: Fixed: the positioner z-index is now applied **only when `backdrop` is set** — backdrop-less drawers keep today's stacking behavior and the documented `--ods-drawer-z-index` keeps working for them. With a backdrop, the variable only reorders inside the positioner; a comment in `_drawer.scss` documents this and points to `positionerStyle`/`backdropStyle` for overrides. (Note: the cssVariable.json table is generated from the compiled CSS, so the entry stays as long as the variable exists — the scss comment is the doc.)

var-poro: With the conditional fix this is now scoped: a backdrop-less drawer keeps its `_overlay.scss` slot (above modal), and a backdrop drawer deliberately adopts modal-par stacking — two modal-like surfaces resolving by DOM order is the same behavior two modals have today.

packages/storybook/stories/components/drawer/drawer.stories.tsx
var-poro: Done: two rendering e2e tests added (backdrop present with the prop, absent by default) — suite passes locally (3/3).

packages/storybook/stories/components/drawer/technical-information.mdx
var-poro: Handled during the rebase onto master (post-#1097): the Backdrop example now lives in `drawer/examples.mdx`, and `technical-information.mdx` keeps the new structure.

var-poro (approved):
Post-rebase re-review: the positioner z-index is now conditional on `backdrop` (no stacking change for existing backdrop-less drawers, documented in `_drawer.scss`), two rendering e2e tests cover backdrop presence/absence (3/3 locally), and the Backdrop example moved to `examples.mdx` per the new structure. Prod Storybook build verified locally. CI green (first run was a flake — identical tree passed on retrigger).

PR #1084, opened 2026-05-26
Add Command component

crotru:


packages/ods-react/src/utils/browser.ts
```diff
@@ -6,4 +8,16 @@ function cssSupports(property: string, value: string): boolean {
...
+  try {
+    return (navigator as NavigatorWithUA).userAgentData?.platform ?? navigator.userAgent ?? '';
+  } catch {
+    return '';
+  }
+}
+
+const platform = getPlatform();
+const isMac = /mac/i.test(platform);
+const isWindows = /win/i.test(platform);
```
var-poro: Still needed?

packages/ods-react/src/components/command/src/components/command-option/CommandOption.tsx
```diff
@@ -0,0 +1,100 @@
...
+
+  useImperativeHandle(ref, () => nodeRef.current!, []);
+
+  useEffect(() => {
+    if (!onSelect || !isVisible || isDisabled) {
+      return;
+    }
+    registerHandler(itemId, onSelect);
+    return () => unregisterHandler(itemId);
+  }, [isDisabled, itemId, onSelect, isVisible, registerHandler, unregisterHandler]);
```
var-poro: with onSelect as a dependency, won't this loop on each render for nothing?

var-poro (changes requested):
missing anatomy image

packages/ods-react/src/components/command/src/contexts/useCommand.tsx
```diff
@@ -0,0 +1,163 @@
...
+  }
+
+  function selectHighlighted(): void {
+    if (highlightedValue) {
+      handlersRef.current.get(highlightedValue)?.();
+    }
+  }
+
+  return (
+    <CommandContext.Provider value={{
```
var-poro: The context `value` object is recreated on every render of the provider, so every consumer — i.e. every `CommandOption` — re-renders on each keystroke / highlight change. Wrapping `value` in `useMemo` (the callbacks are already stable via `useCallback`) would avoid re-rendering the whole list on every interaction.

packages/ods-react/src/components/command/src/contexts/useCommand.tsx
```diff
@@ -0,0 +1,163 @@
...
+
+  function setFilter(value: string): void {
+    setFilterRaw(value);
+  }
+
+  const registerItem = useCallback((value: string, node: HTMLElement): void => {
+    itemNodesRef.current.set(value, node);
+    setRegisteredItems((prev) => {
+      const next = [...prev, value];
+      return next.sort((a, b) => {
```
var-poro: `registerItem` re-sorts the full list with `compareDocumentPosition` on every item registration, so mounting `n` options is ~O(n^2 * log n). Fine for small command palettes, but it can become a bottleneck for large/dynamic lists. Consider sorting once after registration settles, or keying items by mount order.

var-poro (commented):
A few more points in addition to the two you already flagged (dead `isMac`/`isWindows`, and the `onSelect` dependency loop — agreed on both).

Two general notes:
- The PR description is empty — could you add what/why and a short note on the a11y model (combobox + listbox + `aria-activedescendant`)?
- In controlled mode, `Command` passes both `open` and `defaultOpen` to `Modal` (and keeps a local `isOpen` that's then unused). Does `Modal` handle `open` + `defaultOpen` together without warning? Worth confirming or simplifying the controlled/uncontrolled wiring.

Note: the `_modal.scss` body padding change is fine — `--ods-theme-padding-*` is 8px so `* 2` = 16px, identical to the old hardcoded value, and it aligns the body with the header. Only the new `body:first-child` radius rule is worth a quick visual check on header-less modals.

packages/ods-react/src/components/command/src/contexts/useCommand.tsx
```diff
@@ -0,0 +1,163 @@
...
+        return undefined;
+      }
+      if (!registeredItems.includes(current ?? '')) {
+        return registeredItems[0];
+      }
+      return current;
+    });
+  }, [registeredItems]);
+
+  function setFilter(value: string): void {
```
var-poro: Minor: `setFilter` is just a pass-through to `setFilterRaw`. You can expose `setFilterRaw` directly (renamed to `setFilter`) and drop the indirection.

packages/ods-react/src/components/command/src/dev.stories.tsx
```diff
@@ -0,0 +1,341 @@
...
+          {commandItems?.length > 0 &&
+            commandItems.map((group) => (
+              <Command.Group key={group.id} heading={group.label}>
+                {group.items.map((item) => (
+                  <Command.Option key={item.id} onSelect={item.onSelect}>
+                    {item.label}
+                    {item.shortcut &&
+                      <kbd style={{ alignItems: 'center',display: 'flex', gap: '2px' }}>
+                        {getParts(item.shortcut).map((part, idx) => (
+                          <>
```
var-poro: Minor: the `<>` returned inside `.map(...)` has no `key`, which triggers a React key warning. The `key` on `<Kbd>` doesn't count — the root element returned by the map callback needs it. Use `<Fragment key={idx}>` instead of `<>`.

packages/storybook/src/components/themeGenerator/themeGeneratorPreview/ThemeGeneratorPreview.tsx
```diff
@@ -143,6 +145,7 @@ const THEME_PREVIEW_COMPONENTS: ThemePreviewItem[] = [
   { key: 'Checkbox', kind: REACT_COMPONENTS_TITLE.checkbox, label: 'Checkbox' },
   { key: 'Clipboard', kind: REACT_COMPONENTS_TITLE.clipboard, label: 'Clipboard' },
   { key: 'Code', kind: REACT_COMPONENTS_TITLE.code, label: 'Code' },
+  { key: 'Command', kind: REACT_COMPONENTS_TITLE.command, label: 'Command' },
```
var-poro: Minor: alphabetical ordering — `Command` should come after `Combobox` (the next line). Also the import at the top, `import * as Command`, breaks the `XxxStories` naming used by every sibling — rename to `CommandStories`.

packages/ods-react/src/utils/browser.ts
crotru: Not anymore.
I remove

packages/ods-react/src/components/command/src/components/command-option/CommandOption.tsx
crotru: It indeed re-runs each render (but can't loop). 
I'll drop onSelect from the deps via a ref to avoid the needless re-register.

packages/ods-react/src/components/command/src/contexts/useCommand.tsx
crotru: 👍
I also memoize highlight functions

packages/ods-react/src/components/command/src/contexts/useCommand.tsx
crotru: Indeed. I drop the wrapper.

packages/ods-react/src/components/command/src/components/command-group/CommandGroup.tsx
```diff
@@ -0,0 +1,41 @@
...
+}, ref): JSX.Element => {
+
+  return (
+    <div
+      className={ classNames(style['command-group'], className) }
+      data-ods="command-group"
+      ref={ ref }
+      { ...props }
+      role='presentation'
+      aria-label={undefined}>
```
var-poro: Do we really want us not to be able to change the role and aria-label?

packages/ods-react/src/components/command/src/components/command-option/CommandOption.tsx
```diff
@@ -0,0 +1,103 @@
...
+  ...props
+}, ref): JSX.Element | null => {
+  const { filter, highlightItem, highlightedValue, registerHandler, registerItem, unregisterHandler, unregisterItem } = useCommand();
+  const nodeRef = useRef<HTMLDivElement | null>(null);
+  const id = useId();
+  const itemId = props?.id || id;
+  const isDisabled = props['aria-disabled'] === true || props['aria-disabled'] === 'true';
+
+  const isHighlighted = highlightedValue === itemId;
+  const isVisible = !filter || getChildrenText(children).toLowerCase().includes(filter.toLowerCase());
```
var-poro: `getChildrenText(children)` walks the whole children tree on every render — for every option, on every keystroke. Memoizing on `children` won't help (inline JSX = new ref each render). Better to filter on a declared `value`/`label` prop (like combobox `items`) instead of the rendered text: cheaper, and more robust since icons/badges in `children` currently leak into the match.

packages/ods-react/src/components/command/src/components/command-option/CommandOption.tsx
```diff
@@ -0,0 +1,103 @@
...
+    return () => unregisterHandler(itemId);
+  }, [isDisabled, itemId, isVisible, registerHandler, unregisterHandler]);
+
+  useEffect(() => {
+    if (!isVisible || isDisabled || !nodeRef.current) {
+      return;
+    }
+    registerItem(itemId, nodeRef.current);
+    return () => unregisterItem(itemId);
+  }, [isDisabled, isVisible, itemId, registerItem, unregisterItem]);
```
var-poro: Registration is keyed on `isVisible`, so each keystroke unregisters/re-registers the still-matching options. That mutates `registeredItems` in the provider and triggers the three interdependent effects there (highlight reset + reconcile + DOM re-sort via `compareDocumentPosition`) — the whole registry is rebuilt and re-sorted per keystroke. Prefer registering once on mount (by value) and deriving a single `filteredItems` list in the provider (as `useCombobox` does), so typing recomputes one list.

packages/ods-react/src/components/command/src/components/command-empty/CommandEmpty.tsx
```diff
@@ -0,0 +1,32 @@
+import { type ComponentPropsWithRef, type FC, type JSX, forwardRef } from 'react';
+import { useCommand } from '../../contexts/useCommand';
+
+interface CommandEmptyProp extends ComponentPropsWithRef<'div'> {}
+
+const CommandEmpty: FC<CommandEmptyProp> = forwardRef(({
+  children,
+  className,
```
var-poro, line 8: `className` is destructured but never applied to the rendered `div`, so consumer styling is silently dropped.

packages/ods-react/src/components/command/src/components/command-empty/CommandEmpty.tsx
```diff
@@ -0,0 +1,32 @@
...
+  ...props
+}, ref): JSX.Element | null => {
+  const { registeredItems } = useCommand();
+
+  if (registeredItems.length > 0) {
```
var-poro: Disabled options never register (registration is skipped when `aria-disabled`), so if the filter matches only disabled options, the disabled options and the empty state render at the same time. Consider basing this check on visible options rather than navigable ones.

packages/ods-react/src/components/command/src/components/command-content/CommandContent.tsx
```diff
@@ -0,0 +1,98 @@
...
+    return () => el.removeEventListener('animationend', measure);
+  }, [open, nodeRef]);
+
+  return (
+    <ModalContent
+      className={ classNames(style['command-content'], className) }
+      createPortal={ createPortal }
+      data-ods="command-content"
+      dismissible={ false }
+      onKeyDown={ handleKeyDown }
```
var-poro: `onKeyDown` is placed before `{ ...props }`, so a consumer passing their own `onKeyDown` silently disables the whole keyboard navigation. Chain the consumer handler after the internal one, like `CommandOption` does for `onClick`.

packages/ods-react/src/components/command/src/components/command-content/CommandContent.tsx
```diff
@@ -0,0 +1,98 @@
...
+    switch (e.key) {
+      case 'ArrowDown':
+        e.preventDefault();
+        highlightNext();
+        break;
+      case 'ArrowUp':
+        e.preventDefault();
+        highlightPrevious();
+        break;
+      case 'End':
```
var-poro: Home/End are `preventDefault`ed while focus is in the text input, which breaks caret navigation once the user has typed a query. The APG combobox pattern reserves Home/End for the textbox (cmdk does not bind them either). Consider dropping them, or only intercepting when the input is empty.

packages/ods-react/src/components/command/src/components/command-option/CommandOption.tsx
```diff
@@ -0,0 +1,103 @@
...
+      className={ classNames(style['command-option'], isHighlighted && style['command-option--highlighted'], className) }
+      data-ods="command-option"
+      id={ itemId }
+      onClick={ (e) => {
+        if (!isDisabled) {
+          onSelect?.();
+        }
+        props.onClick?.(e);
+      } }
+      onPointerEnter={ () => {
```
var-poro: A consumer-provided `onPointerEnter` is dropped here (internal handler placed after the spread, no chaining), while `onClick` above chains `props.onClick` properly. Chain both for consistency.

packages/ods-react/src/components/command/src/components/command/Command.tsx
```diff
@@ -0,0 +1,35 @@
...
+interface CommandProp extends ModalProp {}
+
+const Command: FC<PropsWithChildren<CommandProp>> = ({
+  children,
+  defaultOpen,
+  onOpenChange,
+  open: openProp,
+  ...modalProps
+}): JSX.Element => {
+  const [isOpen, setIsOpen] = useState(openProp ?? defaultOpen ?? false);
```
var-poro: Nit: seeding internal state from the controlled prop — `useState(defaultOpen ?? false)` is enough, since `openProp ?? isOpen` below already short-circuits in controlled mode.

packages/ods-react/src/components/command/src/components/command-filter/CommandFilter.tsx
```diff
@@ -0,0 +1,45 @@
...
+interface CommandFilterProp extends InputProp {}
+
+const CommandFilter: FC<CommandFilterProp> = forwardRef(({
+  className,
+  onChange,
+  value,
+  ...props
+}, ref): JSX.Element => {
+  const { filter, highlightedValue, id, setFilter } = useCommand();
+  const isControlled = value !== undefined;
```
var-poro, line 23: Passing `value` disables the built-in filtering entirely (the consumer must filter the options themselves). This is intentional and e2e-tested, but documented nowhere — worth a JSDoc note on the prop and a short "controlled filtering" section in the docs, as it is an easy trap.

packages/ods-react/src/components/command/src/components/command-list/CommandList.tsx
```diff
@@ -0,0 +1,37 @@
...
+  children,
+  className,
+  ...props
+}, ref): JSX.Element => {
+  const { highlightedValue, id } = useCommand();
+
+  return (
+    <div className={ classNames(style['command-list__wrapper']) }>
+      <div
+        aria-activedescendant={highlightedValue}
```
var-poro: Nit: `aria-activedescendant` is redundant here — focus stays on the filter input, which already carries it.

packages/storybook/stories/components/command/documentation.mdx
```diff
@@ -0,0 +1,193 @@
...
+<Heading label="Opening & Closing" level={ 3 } />
+
+The **Command** opens as an overlay centered in the viewport.
+
+It can be triggered:
+- Via a dedicated UI trigger
+- Via a keyboard shortcut (⌘ + K on macOS, Ctrl + K on Windows/Linux)
+
+The Command closes when:
+- An item is selected (unless explicitly prevented by the integrator)
```
var-poro: This does not match the implementation: nothing closes the Command on select (see the `WithOnSelect` story — the palette stays open), and there is no built-in way to do it short of fully controlling `open`. Either implement close-on-select (with an opt-out, as is standard for command palettes) or fix the doc.

packages/storybook/stories/components/command/documentation.mdx
```diff
@@ -0,0 +1,193 @@
...
+
+The Command component does not impose constraints on item layout or structure.
+
+<Heading label="Navigation" level={ 2 } />
+
+<Heading label="Focus Management" level={ 3 } />
+
+When the Command opens, focus is set on the search input.
+
+Keyboard navigation moves focus between items in the dropdown.
```
var-poro: Focus does not move between items — it stays on the search input and navigation uses `aria-activedescendant` (as correctly described in the Roles section below). Reword.

packages/storybook/stories/components/command/documentation.mdx
```diff
@@ -0,0 +1,193 @@
...
+
+<Heading label="General Keyboard Shortcuts" level={ 3 } />
+
+Pressing <Kbd>↓</Kbd> or <Kbd>↑</Kbd> moves focus to the next / previous item.
+
+Pressing <Kbd>Enter</Kbd> executes the focused item.
+
+Pressing <Kbd>Escape</Kbd> closes the Command.
+
+Pressing <Kbd>Home</Kbd> moves tothe first item.
```
var-poro: Typo: "moves tothe" → "moves to the".

packages/storybook/stories/components/command/documentation.mdx
```diff
@@ -0,0 +1,193 @@
...
+<Banner of={ CommandStories } />
+
+_The **Command** component provides a fast and efficient way to search, filter, and trigger actions from anywhere in an application._
+
+<Canvas of={ CommandStories.Overview } sourceState="none" />
+
+<Heading label="Overview" level={ 2 } />
+
+<IdentityCard aliases={ ['Command Menu', 'Command Palette', 'Spotlight', 'Quick Actions'] }
+              figmaLink="TODO"
```
var-poro: `figmaLink="TODO"` — to resolve before merge.

var-poro (commented):
Solid PR overall: package structure matches sibling components exactly, the ARIA combobox/listbox/activedescendant pattern is correct, and the e2e coverage is excellent. A few things to address before merge (see inline comments) — the main ones are the dropped `className` in `CommandEmpty` and the docs claiming close-on-select which the component does not do.

Also note the shared `_modal.scss` change: the new body padding computes to the same 16px with the default theme (8px × 2), so it should be safe, but a quick visual pass on existing Modal stories would not hurt since it affects every modal.

packages/ods-react/src/components/command/src/components/command-group/CommandGroup.tsx
crotru: Indeed.
We don't want this element in the accessibility tree (and applying aria-label to an element with role="presentation" would create a role conflict that renders the presentational role ineffective)
We want to keep this tree no matter what user passes:

Image


Component aria-label is passed passed to the `group` role tag one dom-level below

crotru: I add a comment

packages/ods-react/src/components/command/src/components/command-option/CommandOption.tsx
crotru: That would imply a usage as follow:
```tsx
<option label="some string" />
```

But we want to be able to pass full node:
```tsx
<option>Some node</option>
```

Filtering should stay on the rendered children text via getChildrenText, since options take arbitrary nodes and the match should follow what the user actually sees

packages/ods-react/src/components/command/src/components/command-option/CommandOption.tsx
crotru: Done. 
Items now register once and stay registered while filtered out. Typing no longer touches the registry, and the DOM-order re-sort only runs on real mount/unmount.

packages/ods-react/src/components/command/src/components/command-empty/CommandEmpty.tsx
crotru: 😶‍🌫️

crotru: Done

packages/ods-react/src/components/command/src/components/command-empty/CommandEmpty.tsx
crotru: Good catch

packages/ods-react/src/components/command/src/components/command-content/CommandContent.tsx
crotru: 👍

packages/ods-react/src/components/command/src/components/command-content/CommandContent.tsx
crotru: Went with the APG option. Home/End are no longer intercepted and stay with the input caret.

packages/ods-react/src/components/command/src/components/command-option/CommandOption.tsx
crotru: Ok, done.

packages/ods-react/src/components/command/src/components/command/Command.tsx
crotru: 👍

packages/ods-react/src/components/command/src/components/command-filter/CommandFilter.tsx
crotru: Made it more clear and documented it.

packages/ods-react/src/components/command/src/components/command-list/CommandList.tsx
crotru: Removed

packages/storybook/stories/components/command/documentation.mdx
crotru: Doc fixed

packages/storybook/stories/components/command/documentation.mdx
crotru: Home bind removed

packages/storybook/stories/components/command/documentation.mdx
crotru: Wording fixed

packages/storybook/stories/components/command/documentation.mdx
crotru: We don't have a figma for this component

packages/ods-react/src/components/command/src/components/command-group/CommandGroup.tsx
var-poro: Addressed differently: the wrapper has to stay presentational so the listbox only exposes group/option descendants (now explained by a comment in the code), and `aria-label` is properly destructured and carried by the `role="group"` element. Verified — resolving.

packages/ods-react/src/components/command/src/components/command-option/CommandOption.tsx
var-poro: Not memoized, but moot after the registry rework: `children` gets a new JSX identity on every parent render so a `useMemo` keyed on it would never hit, and the costly part (rebuilding the registry on each keystroke) is gone — items now register once and stay registered. Resolving.

var-poro (approved):
Re-reviewed the full PR at c046fa628. All 22 review threads are addressed and verified in code — most behavioral fixes come with a dedicated e2e test (consumer onKeyDown chaining, Home/End caret, disabled-only-match empty state, aria-autocomplete). The registry rework (persistent registration, filteredItems vs navigableItems) is a clean design and the shared gotoStory fix removes a real race for every suite. CI is green.

Two non-blocking notes:
- The public stories now document the namespace form (`Command.Content`…) while sibling components document named imports — assuming this is deliberate since both are exported.
- Merge-order coordination with #1097 still applies (the Demo story / examples.mdx pattern).

PR #1090, opened 2026-06-10
docs: overhaul Storybook documentation structure and homepage

feoche:
Reorganize the entire documentation sidebar into a journey-based
navigation structure, redesign the homepage, and consolidate scattered
pages for a more intuitive developer experience.

## Sidebar restructure

- Introduce journey-based sections: Build, Customize, Resources, Upgrade
- Move guide pages (Accessibility, Charts, i18n, Forms) under Build
- Move Design Tokens, Style Customization, Tailwind under Customize
- Move Code Sandbox under Build, Theme Generator under Customize
- Move LLM Docs, FAQ, and Roadmap under Resources
- Move What's New and migrations under Upgrade
- Update sidebar sort order and SECTION/HOME_TITLE enums accordingly

## Homepage redesign

- Expand Welcome page into a visual showcase with stats and quick paths
- Optimize homepage CSS with ODS tokens, remove dead styles
- Rename "Start Here" to "Get Started", update tagline and descriptions
- Fix homepage card markup (heading presets, semantic spans)
- Hide Welcome page from sidebar (accessible only via logo click)

## Content consolidation

- Consolidate Form Guidelines and Implementing Form into a single Forms page
- Consolidate old migration guides (v12–v18) into Previous Migrations page
- Rename v19 migration title to "Upgrade to v19.x"
- Remove beta badges from What's New section

## Component docs

- Standardize all form component docs to reference the consolidated Forms page
- Fix bugs, typos, and content gaps across documentation pages
- Add pnpm installation option to Get Started page

## Gallery

- Override sbdocs-content max-width on Gallery page for full-width layout
- Add min-height to gallery item preview cards

PR #1097, opened 2026-06-17
feat(core): remove demo pages

crotru:
This PR inspired by (#1090)  cleans up and restructures the Storybook documentation to reduce noise in stories and improve the overall navigation experience.

---

- Remove demo pages from stories.
- Move examples to dedicated MDX pages - every component now has a standalone examples.mdx file containing its usage examples
<kbd>
image
</kbd>

---

- Redesign the homepage with emphasis on responsivity, semantic and a11y.
<kbd>
image
</kbd>
<br />
<br />
<kbd>
image
</kbd>

---

Restructure sidebar navigation - sections reorganized from (Welcome / What's new / Design and Style / Guides / Tools) to (Get Started / Build / Customize / Upgrade / Resources); added a tag-based filter so stories tagged !sidebar are hidden;
- The Welcome page is now only accessible via the logo
<kbd>
image
</kbd>

---

## ⚠️ Breaking: existing deep links

The section renames (`Guides` → `Build`, `Design and Style` → `Customize`, `What's new?` → `Upgrade`, `Form/Guidelines` → `Forms`, …) change the Storybook story IDs. Any previously shared URL (bookmarks, links from the OVH Manager, blog posts, previously published llms files) pointing to a renamed page will land on "story not found". Storybook has no native redirects — this is assumed, and must be mentioned in the release notes.

Note for #1100: the renamed slugs also require regenerating the committed `assets/llms` and updating the changelog exclusion pattern in `postprocess-llms.js` (`what-s-new-changelog` → `upgrade-changelog`).


packages/storybook/.storybook/preview.tsx
```diff
@@ -54,28 +54,36 @@ const preview: Preview = {
...
-              'Accessibility',
-              'Charts - Usage Guidelines',
               '*',
             ],
-            'Tools', [
+            'Resources', [
+              'F.A.Q.',
+              'Sandbox',
+              'Roadmap',
+              'LLM Documentation',
```
var-poro, line 88: As discussed, this might be breaking for LLMs. Would need a double check or simply leave it as it was in AI Agents

var-poro:
How about moving the Recipes in a dedicated section in examples instead of leaving them in Technical information?

packages/storybook/.storybook/preview.tsx
```diff
@@ -54,28 +54,36 @@ const preview: Preview = {
         includeNames: true,
         order: [
           'OVHcloud Design System', [
-            'Welcome',
```
var-poro, line 57: Removing Welcome page means default page is now What's new, not super fund of this

packages/storybook/src/components/homepage/HomepageTile.tsx
```diff
@@ -0,0 +1,40 @@
...
+      }}>
+        <div className={styles["homepage-tile__title"]}>
+          <Icon className={styles["homepage-tile__icon"]} name={icon} />
+          <Text as="h2" preset={TEXT_PRESET.heading6}>
+            {title}
+          </Text>
+        </div>
+        <Text>{description}</Text>
+      </a>
+    </Card>
```
var-poro, line 36: Would be more cohesive to use ODS Links instead of a non-approved behavior we have in don'ts (to use a button as a link)

packages/storybook/src/components/homepage/Homepage.tsx
```diff
@@ -1,37 +1,99 @@
...
+          </Link>
 
           <ExternalLink href="https://zeroheight.com/6fc8a63f7/p/533db0-ovhcloud-design-system">
             Design Guidelines
           </ExternalLink>
 
           <ExternalLink href="https://github.com/ovh/design-system">
+            <Icon name={ICON_NAME.github} />
             GitHub repository
           </ExternalLink>
```
var-poro: Might be completely subjective, but I feel like having all the links in the same row makes it harder on the first glance to understand it's not one very long link and 3 distincts one

packages/storybook/.storybook/preview.tsx
```diff
@@ -54,28 +54,36 @@ const preview: Preview = {
         includeNames: true,
         order: [
```
var-poro, line 55: TBD with entire team

var-poro (changes requested):

packages/storybook/.storybook/preview.tsx
crotru: Actually "Welcome" is an entry that is already hidden in css.
I shall add it back here. 
Then we'll have this intented behavior :
Welcome is the default page, it is not visible in the sidebar, this is the page you reach when clicking on the logo

packages/storybook/src/components/homepage/HomepageTile.tsx
crotru: Agreed.

What do you think of this:

image

packages/storybook/src/components/homepage/Homepage.tsx
crotru: This is organized from top to bottom on smaller screen.
I make the change to have the same on larger screens

packages/storybook/.storybook/preview.tsx
crotru: Changed

astagnol (changes requested):
There is an issue with badges in dark theme:
Image

astagnol (approved):

var-poro (commented):
The mechanical part of this PR is impeccable — I cross-checked all 60 components by script: every `<Canvas of={ Stories.X }>` in the new `examples.mdx` files resolves to a real export, no dead imports left in the `technical-information.mdx` files, no leftover `Demo` references anywhere, and `ThemeGeneratorPreview` uses the `ThemeGenerator` story so nothing breaks there. Generator templates and CONTRIBUTING.md updated consistently, and the chat recipe scroll fix is correct.

The real topics are cross-cutting (see inline comments):
- the section renames break every existing deep link (Storybook story IDs derive from titles) — probably assumed, but worth stating explicitly in the PR and at release time;
- merge-order coordination with #1100 (all committed `assets/llms` become stale + the changelog exclusion pattern stops matching) and with #1084 (Command still ships a `Demo` story and no `examples.mdx` — it will need a rebase onto this structure).

Also note both this PR and #1100 carry the identical `generate-version-list.js` change — harmless, but one of them can drop it.

packages/storybook/src/constants/meta.ts
```diff
@@ -3,35 +3,47 @@
 // Until a solution is found, those are still useful when defining <StorybookLink />
 
 enum SECTION {
-  design = 'Design and Style',
-  guides = 'Guides',
+  build = 'Build',
```
var-poro, line 6: Heads-up: these section renames (`Guides` → `Build`, `Design and Style` → `Customize`, `What's new?` → `Upgrade`, `Form/Guidelines` → `Forms`, …) change the Storybook story IDs, so every existing external deep link (bookmarks, links from the OVH Manager, blog posts, previously published llms files) will land on "story not found". Storybook has no native redirects — if this is assumed, it should at least be called out in the PR description and in the release notes.

packages/storybook/stories/ovhcloud-design-system/whatsnew/changelog.mdx
```diff
@@ -2,7 +2,7 @@ import { Markdown, Meta } from '@storybook/blocks';
 import ChangeLog from '../../../../../CHANGELOG.md?raw';
 import { NoToC } from '../../../src/components/noToC/NoToC';
 
-<Meta title="OVHcloud Design System/What's new?/Changelog" />
+<Meta title="OVHcloud Design System/Upgrade/Changelog" />
```
var-poro, line 5: Coordination with #1100: this slug change (`what-s-new-changelog` → `upgrade-changelog`) means the exclusion pattern in `postprocess-llms.js` (`/^ovhcloud-design-system-what-s-new-changelog/`) no longer matches, so the changelog — excluded on purpose to avoid bloat — will silently re-enter `llms-full.txt`. More broadly, all the renamed pages invalidate the committed `assets/llms`: whichever PR merges second must regenerate them and update the patterns.

packages/storybook/src/components/externalLink/ExternalLink.tsx
```diff
@@ -23,7 +23,9 @@ const ExternalLink = ({ children, href, ...prop }: LinkProp & Omit<ComponentProp
     <Link
       { ...prop }
       href={ resolvedHref }
-      target="_blank">
+      size="small"
```
var-poro: `size="small"` is hardcoded after the spread, so it changes every `ExternalLink` on every doc page and callers can no longer override it. If this is meant for the homepage, pass the prop at the call site instead.

packages/storybook/assets/css/storybook.css
```diff
@@ -26,17 +26,22 @@ button[title="Shortcuts"] {
...
-  text-transform: none;
 }
 
 .sidebar-header {
   display: grid !important;
   grid-template-columns: max-content 1fr max-content;
 }
 
+/* Hide Welcome page from sidebar (accessible via logo click) */
+.sidebar-item[data-item-id="ovhcloud-design-system-welcome--docs"] {
```
var-poro, line 43: The Welcome page is now hidden twice: this CSS rule and the `!sidebar` tag + manager filter. If the filter works, this selector is dead (and fragile — it hardcodes a story ID); if it doesn't, the tag is. Keep one mechanism.

packages/storybook/src/components/homepage/homepage.module.css
```diff
@@ -1,19 +1,80 @@
...
+}
+
+.homepage__divider {
+  inline-size: 100%;
+}
+
+.homepage__tiles {
+  display: flex;
+  flex-wrap: wrap;
+  grid-template-columns: repeat(3, 1fr);
```
var-poro: `grid-template-columns` has no effect on a flex container — dead declaration (the block uses `display: flex`).

packages/storybook/src/components/homepage/homepage.module.css
```diff
@@ -1,19 +1,80 @@
...
+  gap: 0.75rem;
+  margin-block: 1rem;
+  max-inline-size: 60rem;
+}
+
+.homepage__cta {
```
var-poro: `.homepage__cta` is not used anywhere — dead CSS. Also, this file and `storybook.css` lost their trailing newline.

packages/storybook/src/components/homepage/Homepage.tsx
```diff
@@ -1,45 +1,105 @@
...
+  ICON_NAME,
+  TEXT_PRESET,
+  Icon,
+  Text,
+  Divider,
+  DIVIDER_SPACING,
+  Badge,
+  Link,
+  BADGE_COLOR,
+} from "@ovhcloud/ods-react";
```
var-poro: Nit: the homepage files (`Homepage.tsx`, `Brand.tsx`, `HomepageTile.tsx`) switched to Prettier-style formatting (double quotes, `{prop}` without spaces) while the rest of the repo uses single quotes and `{ prop }`. The storybook package has no lint script so nothing catches it — worth realigning manually.

packages/storybook/src/constants/meta.ts
```diff
@@ -3,35 +3,47 @@
...
-  i18nGuide = `${SECTION.home}/${SECTION.guides}/Internationalization`,
-  implementingFormGuide = `${SECTION.home}/${SECTION.guides}/Form/Implementing a Form`,
-  migration12to13 = `${SECTION.home}/${SECTION.whatsNew}/${SECTION.migrationGuide}/12.x to 13.x`,
-  migration13to14 = `${SECTION.home}/${SECTION.whatsNew}/${SECTION.migrationGuide}/13.x to 14.x`,
-  migration14to15 = `${SECTION.home}/${SECTION.whatsNew}/${SECTION.migrationGuide}/14.x to 15.x`,
-  migration15to16 = `${SECTION.home}/${SECTION.whatsNew}/${SECTION.migrationGuide}/15.x to 16.x`,
-  migration16to17 = `${SECTION.home}/${SECTION.whatsNew}/${SECTION.migrationGuide}/16.x to 17.x`,
-  migration17to18 = `${SECTION.home}/${SECTION.whatsNew}/${SECTION.migrationGuide}/17.x to 18.x`,
+  i18nGuide = `${SECTION.home}/${SECTION.build}/Internationalization`,
+  implementingFormGuide = `${SECTION.home}/${SECTION.build}/Forms`,
```
var-poro: Nit: `formGuidelines` and `implementingFormGuide` are now two enum entries with the same value (the guide merge itself is clean — all old links resolve). One of them can go.

packages/storybook/src/constants/meta.ts
var-poro: Documented in the PR description (⚠️ Breaking: existing deep links) — to be echoed in the release notes.

packages/storybook/stories/ovhcloud-design-system/whatsnew/changelog.mdx
var-poro: Tracked: #1100 will regenerate the committed assets and update the exclusion pattern after this PR merges (also noted in this PR's description).

packages/storybook/src/components/externalLink/ExternalLink.tsx
var-poro: Fixed: `size="small"` moved before the spread — it is now an overridable default instead of a forced value.

packages/storybook/assets/css/storybook.css
var-poro: Verified at runtime: tags set on unattached MDX do not reach the story index in Storybook 8.6, so the `!sidebar` filter never sees the Welcome entry — the CSS is what actually hides it (checked via puppeteer: item present in the sidebar DOM, hidden by CSS). Both mechanisms are kept on purpose: the tag/filter is the target mechanism for after the Storybook upgrade; the CSS comment now documents this.

packages/storybook/src/components/homepage/homepage.module.css
var-poro: Fixed: dead declaration removed.

packages/storybook/src/components/homepage/homepage.module.css
var-poro: Fixed: dead class removed, trailing newlines restored.

packages/storybook/src/components/homepage/Homepage.tsx
var-poro: Fixed: the three homepage files are realigned on the repo style (single quotes, spaced JSX expressions, multiline export).

packages/storybook/src/constants/meta.ts
var-poro: Fixed: `implementingFormGuide` removed, its usages now point to `formGuidelines`.

var-poro (approved):
Re-reviewed post-rebase: Command migrated to the examples pattern (rendering verified in a live Storybook — sidebar clean, examples page renders its 8 canvases without error, homepage OK), all review threads addressed, deep-links breakage documented in the description. CI green.

PR #1100, opened 2026-06-25
feat(llms): Improve llms.txt

var-poro:


crotru (approved):

astagnol (approved):

packages/storybook/scripts/postprocess-llms.js
```diff
@@ -144,7 +287,24 @@ function buildFullDoc(allDocs) {
...
+    .map((word) => word.charAt(0).toUpperCase() + word.slice(1))
+    .join(' ');
+}
+
+// Relative link used between llms files. It resolves to the sibling file both
+// locally (integrators reading the files shipped in their node_modules) and on
+// the web (within the versioned /v<version>/llms/ path), so links stay
+// version-correct everywhere without hard-coding an absolute host.
+function relUrl(file) {
+  return `./${file}`;
```
var-poro: The switch from absolute to relative links breaks the **dist-root copies** of the navigation files. The script writes the aggregates/indexes both at the `dist/` root and in `dist/llms/`, but the individual files (`react-components-button.txt`, …) only exist in `dist/llms/`. An agent entering through the root `/llms.txt` — the canonical llmstxt.org location — follows `./ods-components-index.txt` (exists at root) then `./react-components-button.txt` → 404. Same problem for the root `llms-index.json`.

Options: make the root copies link into `./llms/…`, also write the individual files at the root, or stop publishing the indexes at the root. Note that `verify-llms.sh` can't catch this: it only checks `assets/llms/`, where everything is flat.

packages/storybook/package.json
```diff
@@ -13,6 +13,7 @@
   "scripts": {
     "build:storybook": "pnpm generate:versions && storybook build -o dist",
     "extract:llms": "storybook-llms-extractor --config llms.config.js && node scripts/postprocess-llms.js",
+    "verify:llms": "bash scripts/verify-llms.sh",
```
var-poro, line 16: Is `verify:llms` wired into the internal CI / release pipeline? Nothing in the repo runs it, and its guarantees (determinism, version consistency) only hold if it runs systematically. The release flow is where it matters most: the front-matter embeds the release version and the versioned canonical URL, so every release must regenerate the assets or `ods-react` ships docs with a stale version.

packages/storybook/scripts/verify-llms.sh
```diff
@@ -0,0 +1,176 @@
...
+ASSETS="$ROOT/packages/storybook/assets/llms"
+DIST_REACT="$ROOT/packages/ods-react/dist/llms"
+VERSION="$(node -e "console.log(require('$ROOT/lerna.json').version)")"
+FULL="${1:-}"
+
+GENERATED="llms.txt llms-full.txt ods-documentation-components.txt ods-documentation-generic.txt ods-components-index.txt ods-generic-index.txt llms-index.json"
+NAV_FILES="llms.txt ods-components-index.txt ods-generic-index.txt"
+
+pass=0; fail=0
+ok()      { printf '  \033[32m✓\033[0m %s\n' "$1"; pass=$((pass + 1)); }
```
var-poro, line 26: Nit: the script's output messages are in French while every other script in this public repo (including `postprocess-llms.js`) logs in English — worth harmonizing.

packages/storybook/scripts/verify-llms.sh
```diff
@@ -0,0 +1,176 @@
...
+nav_bad=0
+for b in $NAV_FILES; do
+  grep -qE '\]\(https?://' "$ASSETS/$b" && nav_bad=$((nav_bad + 1))
+done
+[ "$nav_bad" = "0" ] \
+  && ok "liens de navigation relatifs (aucun http(s) dans llms.txt / index)" \
+  || ko "$nav_bad fichier(s) de navigation avec lien web absolu"
+
+# Front-matter source stays absolute & versioned (canonical citation)
+sample="$(individual_files | head -1)"
```
var-poro: The absolute/versioned `source:` pattern is only checked on the first individual file — looping over all of them costs nothing and catches a partial regeneration.

patches/@fluentui__storybook-llms-extractor@0.0.3.patch
```diff
@@ -64,3 +64,66 @@ index 2852bb0e7f7763a7ba2d0b6765554921917f3d91..2c7b0f50978c3dfc2864c84dc4e85771
...
++        });
+         await page.waitForSelector('.sbdocs-content', {
+             state: 'attached',
+-            timeout: 2000
++            timeout: 30000
+         });
+-        const html = await page.locator('.sbdocs-content').innerHTML();
++        let html = '';
++        let stable = 0;
++        for(let i = 0; i < 40 && stable < 3; i++){
```
var-poro, line 117: If the DOM never stabilizes, the loop silently captures the last HTML after 40 iterations. A `console.warn` when `stable < 3` on loop exit would make a future flaky extraction much easier to diagnose.

packages/storybook/scripts/postprocess-llms.js
```diff
@@ -3,14 +3,18 @@
...
-// const { version } = require('../../../lerna.json');
-const version = '19.7.0';
+const { version } = require('../../../lerna.json');
 
 const DIST_DIR = path.join(__dirname, '..', 'dist');
 const SUMMARY_FILENAME = 'llms.txt';
 const SUMMARY_PATH = path.join(DIST_DIR, SUMMARY_FILENAME);
 const LLMS_DIR = path.join(DIST_DIR, 'llms');
+const ASSETS_LLMS_DIR = path.join(__dirname, '..', 'assets', 'llms');
+const DIST_LLMS_DIR = LLMS_DIR;
```
var-poro: Nit: `DIST_LLMS_DIR` is a redundant alias of `LLMS_DIR`, and `GENERIC_CATEGORY.pattern` is never used (`categorizeDocs` only tests the components pattern) — micro-cleanup.

var-poro (commented):
Strong PR overall: the self-describing front-matter, token estimates, machine-readable index and local distribution in `ods-react/dist/llms` form a coherent design, the idempotence is real (generated files excluded from re-ingestion, stale aggregates pruned — re-running the post-process is a no-op), and the determinism patch (frozen clock, seeded PRNG, settled-DOM polling) is well thought out.

One real bug to fix before merge (broken relative links in the dist-root navigation copies) and one open question (is `verify:llms` wired into the internal CI / release pipeline?) — see inline comments.

packages/storybook/scripts/postprocess-llms.js
var-poro: Fixed: navigation files now exist in two variants — the dist-root copies link into `./llms/…` while the flat-directory copies (dist/llms, assets, node_modules) link to their siblings. `verify-llms.sh` also gained a check asserting every relative navigation link resolves to an existing sibling, so this bug class can't come back silently.

packages/storybook/scripts/verify-llms.sh
var-poro: Fixed: all output messages are in English now.

packages/storybook/scripts/verify-llms.sh
var-poro: Fixed: the `source:` check now loops over every individual file.

patches/@fluentui__storybook-llms-extractor@0.0.3.patch
var-poro: Fixed: the patch now emits a `console.warn` when the DOM never settles (40 iterations without 3 stable reads) instead of silently capturing the last snapshot.

packages/storybook/scripts/postprocess-llms.js
var-poro: Fixed: both removed.

packages/storybook/package.json
var-poro: The internal pipeline should already cover this — resolving. Worth a double-check on the next release run.

PR #1101, opened 2026-06-29
chore(deps): toolchain upgrade - ESLint, TS, Jest, Husky, Stylelint bumps

crotru:
## What

Full toolchain upgrade across the monorepo:

- **ESLint 8 → 10** with flat config migration (`eslint.config.mjs` replaces `.eslintrc`), **`@typescript-eslint`** 7 → 8
- **TypeScript 5 → 6**
- **Husky 4 → 9** — hook files in `.husky/`, `prepare` script wired up
- **Jest 29 → 30** — pnpm override added for `jest-environment-node` (jest-puppeteer still pins v29)
- **Stylelint 16 → 17** — pnpm override added for `stylelint-order` (concentric-order still pins v7)
- **Puppeteer 22 → 25**, **Shiki 3 → 4**, **Vite plugins**, and misc minor bumps
- **Example apps**: React 19.2.7, Vite 8, Next.js 16, react-router-dom 7, Babel 8, updated test deps

CSS fix: replaced `align-items/justify-content: start/end` with `flex-start/flex-end` on flex containers (autoprefixer warning when consumed by Next.js)
Removed dead `test:wrapper:react` / `test:wrapper:vue` root scripts

## Intentionally not upgraded

| Package | Reason |
|---|---|
| `storybook` | Separate migration (8 → 10), not in scope |
| `vite` (ods-react / ods-recipes) | Blocked by Storybook 8 — `@storybook/react-vite` requires Vite ≤ 6 |
| `react` & `@types/react` (ods-react / ods-recipes) | Not upgraded outside example apps |
| `tailwindcss` | v4 is a full rewrite, requires dedicated migration |

---

## Post-review updates

- Dev stories are linted again (dedicated flat-config block, relaxed rules + the `no-restricted-imports` barrel guard restored); only `**/tests/**/*.stories.tsx` stay ignored, as before. NB: per-package `lint:ts` scripts still skip stories via `--ignore-pattern`, unchanged from the previous setup.
- `test:wrapper:react` / `test:wrapper:vue` root scripts are now actually removed.
- `docgen.ts` unescapes typedoc-escaped JSDoc (`\@`, `\<`, …) so the technical specification tables render clean (needed for the escaped `@type=` in `Markdown.tsx`).
- `func-call-spacing` restored (still shipped by ESLint 10).
- `eslint-plugin-tsdoc` dropped (its only rule was never enabled).
- API note: `Code.highlighter.language`/`theme` become optional (consistent with the runtime guards — `highlighter={{}}` is now valid typed input).
- TypeScript 6 adaptation: `CommandTriggerProp` now extends `ModalTriggerProp` (TS6 flags the previous `ComponentPropsWithRef<'button'>` spread against the narrowed `value` prop).
- Follow-up issue for the `react-hooks/set-state-in-effect` file-wide disables: #1104


eslint.config.mjs
```diff
@@ -0,0 +1,168 @@
...
+import jsxA11yPlugin from 'eslint-plugin-jsx-a11y';
+import reactPlugin from 'eslint-plugin-react';
+import reactHooksPlugin from 'eslint-plugin-react-hooks';
+import tsdocPlugin from 'eslint-plugin-tsdoc';
+
+export default [
+  {
+    ignores: [
+      'packages/storybook/**',
+      '**/*.stories.tsx',
```
var-poro: The old `.eslintrc` only ignored `**/tests/**/*.stories.tsx` and had a dedicated override for `**/src/dev.stories.tsx`: relaxed rules, **but** with `no-restricted-imports` banning barrel imports of `@ovhcloud/ods-react` (with an explicit message pointing to relative imports). This blanket `**/*.stories.tsx` ignore drops both: dev stories are no longer linted at all, and the barrel guard silently disappears. Suggest restoring a dedicated block for `**/src/dev.stories.tsx`.

package.json
```diff
@@ -34,7 +34,7 @@
     "ods:publish": "lerna publish --concurrency 1 from-package --yes",
     "storybook": "pnpm build:storybook:deps && pnpm --filter @ovhcloud/ods-storybook run start",
     "test:e2e:ci": "lerna run --stream --concurrency 1 test:e2e:ci",
-    "test:spec:ci": "lerna run --stream --concurrency 1 test:spec:ci",
+    "test:spec:ci": "lerna run --stream --concurrency 2 test:spec:ci",
     "test:wrapper:react": "lerna run --stream --concurrency 1 test:wrapper:react",
```
var-poro: The PR description says the dead `test:wrapper:react` / `test:wrapper:vue` root scripts were removed, but they're still here.

packages/ods-react/src/components/markdown/src/components/markdown/Markdown.tsx
```diff
@@ -10,7 +10,7 @@ import { MarkdownCode, MarkdownPre } from '../markdown-code/MarkdownCode';
 
 interface MarkdownProp {
   /**
-   * @type=Record<keyof JSX.IntrinsicElements, ReactNode>
+   * \@type=Record\<keyof JSX.IntrinsicElements, ReactNode\>
```
var-poro, line 13: This custom `@type=` tag is parsed by `packages/storybook/src/helpers/docgen.ts` with a regex over the raw comment text. It will still match `@type`, but it captures the value **with the backslashes**, so the Technical Information table will likely render `Record\<keyof JSX.IntrinsicElements, ReactNode\>` literally. If typedoc 0.28 requires the escaping, the unescaping should happen in `docgen.ts` — worth checking the rendered Markdown props table either way.

packages/ods-react/src/components/code/src/components/code/Code.tsx
```diff
@@ -24,11 +24,11 @@ interface CodeProp extends ComponentPropsWithRef<'div'> {
     /**
      * The programming language displayed.
      */
-    language: LanguageRegistration[],
+    language?: LanguageRegistration[],
```
var-poro, line 27: Making `language`/`theme` optional is a public typing change (a loosening — `highlighter={{}}` becomes valid input) slipped into a chore PR. It's consistent with the runtime guards so probably fine, but it should at least be mentioned in the PR description.

eslint.config.mjs
```diff
@@ -0,0 +1,168 @@
...
+      '@typescript-eslint/no-wrapper-object-types': 'error',
+      'array-bracket-spacing': ['error', 'never'],
+      'arrow-parens': ['error', 'always'],
+      'arrow-spacing': ['error', { after: true, before: true }],
+      'brace-style': ['error', '1tbs'],
+      'comma-dangle': ['error', 'always-multiline'],
+      'complexity': ['error'],
+      'curly': ['error', 'all'],
+      'default-case': 'error',
+      'func-style': ['error', 'declaration', { allowArrowFunctions: true }],
```
var-poro, line 78: `func-call-spacing` (present in the old config) is gone, while the rest of the same deprecated-formatting-rules batch (`indent`, `quotes`, `semi`, …) was kept — intentional drop or lost in the migration?

pnpm-workspace.yaml
```diff
@@ -8,16 +8,40 @@ packages:
...
 allowBuilds:
+  '@parcel/watcher': false
   esbuild: true
   husky: true
   nx: true
   puppeteer: true
   sharp: true
+  unrs-resolver: true
+
+minimumReleaseAgeExclude:
```
var-poro, line 21: Nothing in the repo sets `minimumReleaseAge` (neither here nor in an `.npmrc`). If pnpm 11 doesn't enable it by default, this exclude list is dead config — and if it does, a one-line comment stating the effective delay would help.

eslint.config.mjs
```diff
@@ -0,0 +1,168 @@
...
+          allowSeparatedGroups: false,
+          ignoreCase: false,
+          ignoreDeclarationSort: true,
+          ignoreMemberSort: false,
+          memberSyntaxSortOrder: ['none', 'all', 'multiple', 'single'],
+        },
+      ],
+      'sort-keys': ['error', 'asc'],
+      'space-before-function-paren': ['error', 'never'],
+      'tsdoc/syntax': 'off',
```
var-poro: Nit: `eslint-plugin-tsdoc` is installed and registered but its only rule is `off` (this was already true in the old config — the rule was never enabled). Might as well drop the plugin.

.husky/commit-msg
```diff
@@ -0,0 +1 @@
+pnpm exec commitlint --edit $1
```
var-poro: Nit: quote it — `pnpm exec commitlint --edit "$1"` — a checkout path with a space would break the hook.

packages/ods-react/src/components/combobox/src/contexts/useCombobox.tsx
```diff
@@ -1,3 +1,5 @@
+/* eslint-disable react-hooks/set-state-in-effect */
```
var-poro, line 1: The file-wide `react-hooks/set-state-in-effect` disables (here, prompt-input, query-filter) are the right call for a deps PR — no behavior change — but the rule flags real patterns worth revisiting. A follow-up issue/TODO would keep them from becoming permanent.

var-poro (commented):
High-quality PR for a bump of this size. Highlights: the `pnpm-workspace.yaml` overrides are exemplary (each one documents the exact cause and a REMOVE WHEN condition), the `ban-types` migration is faithful (`no-empty-object-type` with `with-single-extends` preserves the `interface XProp extends YProp {}` pattern, plus `no-wrapper-object-types`), and the e2e changes strengthen the tests against flakiness (polling via `waitForFunction`) without weakening any expected state.

Three things to fix before merge (see inline comments): dev stories are no longer linted at all and the barrel-import guard silently disappeared; the PR description claims `test:wrapper:*` scripts were removed but they're still there; and the escaped JSDoc in `Markdown.tsx` likely leaks literal backslashes into the Technical Information props table.

eslint.config.mjs
var-poro: Fixed: the blanket ignore is narrowed back to `**/tests/**/*.stories.tsx` and a dedicated dev-stories block restores the relaxed rules + the `no-restricted-imports` barrel guard. NB: the per-package `lint:ts` scripts still skip stories via `--ignore-pattern` — that was already the case before, the override guards IDE/direct runs like the old `.eslintrc` did.

package.json
var-poro: Fixed: both scripts are now actually removed.

packages/ods-react/src/components/markdown/src/components/markdown/Markdown.tsx
var-poro: Fixed in `docgen.ts`: JSDoc escapes (`\\@`, `\\<`, …) are stripped before tag parsing and display, so the table renders `Record<keyof JSX.IntrinsicElements, ReactNode>` clean. Verified with a prod Storybook build.

packages/ods-react/src/components/code/src/components/code/Code.tsx
var-poro: Now mentioned in the PR description (Post-review updates section).

eslint.config.mjs
var-poro: Checked: the rule is still shipped by ESLint 10 — restored for parity.

pnpm-workspace.yaml
var-poro: `pnpm config get minimumReleaseAge` is unset and pnpm 11 does not enable it by default here. The list is kept (it protects an eventual CI/global activation from blocking these pinned versions) with a comment stating exactly that.

eslint.config.mjs
var-poro: Dropped (plugin + devDependency).

.husky/commit-msg
var-poro: Fixed: `"$1"` quoted.

packages/ods-react/src/components/combobox/src/contexts/useCombobox.tsx
var-poro: Follow-up issue created: #1104.

var-poro (approved):
Post-rebase re-review. All 9 review threads addressed (dev-stories lint + barrel guard restored — and actually enforced in CI now since flat-config ignore patterns changed semantics; docgen unescapes typedoc-escaped JSDoc; func-call-spacing restored; follow-up issue #1104). The rebase also surfaced real incompatibilities between the new toolchain and recently merged code, fixed here: CommandTrigger typing under TS6, command lint errors under ESLint 10 / react-hooks 7, concentric order in command's scss. Verified locally end to end: full lint, tsc -b, prod Storybook build, jest 30 spec suites, command (61/61) + drawer (3/3) e2e under puppeteer 25, and all five example app builds. CI green.

PR #1107, opened 2026-08-25
fix(select): pluralize the merged selection default label

feoche:
The default label for a merged multiple selection was the fixed string 'Selected item', which read incorrectly for more than one selected item. Derive it from the total instead, and fall back to it when the consumer passes an empty label as well as when the prop is omitted.

### DIFF
--- a/packages/ods-react/src/components/select/src/components/select-merged-selection/SelectMergedSelection.tsx
+++ b/packages/ods-react/src/components/select/src/components/select-merged-selection/SelectMergedSelection.tsx
@@ -10,12 +10,14 @@ interface SelectMergedSelectionProp {
 
 const SelectMergedSelection: FC<SelectMergedSelectionProp> = ({
   disabled,
-  multipleSelectionLabel = 'Selected item',
+  multipleSelectionLabel,
   total,
 }): JSX.Element => {
+  const label = multipleSelectionLabel || `Selected item${total > 1 ? 's' : ''}`;
+
   return (
     <span>
-      { multipleSelectionLabel }&nbsp;
+      { label }&nbsp;
       <span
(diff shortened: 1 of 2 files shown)


packages/ods-react/src/components/select/src/components/select-merged-selection/SelectMergedSelection.tsx
```diff
@@ -10,12 +10,14 @@ interface SelectMergedSelectionProp {
 
 const SelectMergedSelection: FC<SelectMergedSelectionProp> = ({
   disabled,
-  multipleSelectionLabel = 'Selected item',
+  multipleSelectionLabel,
   total,
 }): JSX.Element => {
+  const label = multipleSelectionLabel || `Selected item${total > 1 ? 's' : ''}`;
```
var-poro, line 16: Switching to || means multipleSelectionLabel="" now shows the default instead of the counter alone, is this intended?

packages/ods-react/src/components/select/tests/component/selectMergedSelection.spec.tsx
var-poro, line 1: Please align with the existing test layout (tests/controller/) instead of introducing a new tests/component/ + renderToString pattern.

packages/ods-react/src/components/select/src/components/select-merged-selection/SelectMergedSelection.tsx
var-poro, line 1: Instead of adding an English exclusive plural rule, wouldn't it be better to add SELECT_I18N just like the other components that handle translation?

var-poro (changes requested):

PR #1108, opened 2026-08-26
feat(docs)!: replace the Storybook site with the ODS docs platform

var-poro:
## Summary

Replaces the Storybook documentation site with a custom docs platform built on ODS components, per the July scoping (public docs only; the per-component e2e storybooks in ods-react are untouched — phase 2).

**Past versions are not affected**: every published `@ovhcloud/ods-storybook` version keeps its frozen build on npm and keeps being served under its `/vX.Y.Z/` URL forever. This PR only changes what ships from the next release on.

## What the platform provides

- All **58 components**: migrated documentation (neutral MDX, no imports), live demos rendered from the real public CSF stories via portable-stories in isolated iframes, an Examples tab, a Technical Information tab generated from the typedoc JSON (props, enums, CSS variables, interactive anatomy), per-component recipes.
- All guides, helpers, recipes, changelog; **icon gallery** (searchable, 500+ icons); **theme generator** (token editors, palette generation, JSON import/export, live preview of every component); **code sandbox** (Monaco + full ODS type graph, shareable snippets); ⌘K search with grouped results; dark mode; a11y basics (skip link, landmarks, reduced motion).
- **llms output emitted natively at build time** from the same sources the pages render — no DOM scraping. Same file naming and front-matter contract as before (verify-llms.sh: green), committed under packages/docs/assets/llms as before, shipped in the ods-react npm package, now covering all 58 components (264 files). A root llms.txt lists every deployed version.
- React 19-native, Vite 8. No Storybook dependency for the public docs.

## Deployment continuity

- gh-pages: same script (`generate-gh-pages.mjs`), now walking a **dual lineage** — frozen Storybook builds from `@ovhcloud/ods-storybook`, new versions from `@ovhcloud/ods-docs`. Deep links work on gh-pages through a root 404 bounce. Also fixes a latent re-deploy bug (manual deploy on a released master nested the fresh build inside the npm copy).
- CI: **no CDS change required** — the docs package answers `build:storybook`/`generate:versions` by script name and mirrors its dist into the path the branch preview serves (`packages/storybook/dist`). Updating the preview `source:` line in ods-github-cds becomes optional cleanup.
- First release after merge publishes `@ovhcloud/ods-docs` and deploys the new platform as vNEXT + latest.

## Validation

- Repo test harness (packages/docs/tests, `pnpm test:smoke`): gh-pages-semantics server, 203-route sweep, theme generator, icon gallery, snippet audit — all green.
- Every "open in sandbox" snippet compiles and renders (423/423, browser-replayed).
- Full gh-pages rehearsal against real npm tarballs: 78 past versions deployed and booting next to the new platform.
- verify-llms.sh 12/12 including the ods-react local distribution; e2e/lint/unit suites green (CDS run: success on this branch).

Includes the merged drawer pointer-events fix (#1106) and completes 26 story import headers surfaced by the snippet audit.


pnpm-workspace.yaml
crotru: This can go 🧹

crotru:
⚠️ Claude spotted this:
Quite concerning imho

---

### Deploy lineage: `19.7.3` can land in both lineages, and the loop isn't idempotent on version collisions

Two small things in `scripts/generate-gh-pages.mjs`, both around the dual-lineage change. The mechanism is right, I think it just needs one guard rail.

**1. `@ovhcloud/ods-docs` is currently versioned `19.7.3`, which the Storybook lineage already owns.**

The comment at the top of the script draws the boundary clearly:

```js
// Dual lineage: versions <= 19.7.x are frozen Storybook builds published as
// @ovhcloud/ods-storybook; newer versions are the docs platform.
```

But `packages/docs/package.json` is at `19.7.3`, and `@ovhcloud/ods-storybook@19.7.3` is already on npm (it's the current `lerna.json` version). `@ovhcloud/ods-docs` isn't published at all yet. Since `ods:publish` is `lerna publish from-package`, which publishes anything whose version isn't on the registry, a publish run from master in its current state would push `@ovhcloud/ods-docs@19.7.3` and put `19.7.3` in both lineages, contradicting the boundary above.

In the happy path this never happens: the `!` in this PR's title means the next `lerna version --conventional-commits` computes a major, so the first docs release lands at `20.0.0` and there's no overlap at all. The window is narrow, it's just also cheap to close, and the script currently has no defence if it ever opens.

**2. The lineage loop nests instead of replacing on a collision.**

```js
await $`mv ${dir}/dist dist/v${version}`;
```

`packageNames` is iterated Storybook-first, so on a duplicate version the Storybook tarball creates `dist/v19.7.3`, then the docs tarball's `mv` lands *inside* it as `dist/v19.7.3/dist`. This is exactly the failure you already diagnosed for the current build a few lines down:

```js
// ... The npm loop may already have deployed this same version (manual
// re-deploy on a released master): replace it, or cp would nest into the
// existing dir.
await $`rm -rf dist/v${currentVersion}`;
```

Right now that block masks the symptom, because `currentVersion` happens to be the colliding version and it overwrites `dist/v19.7.3` wholesale. Once master moves past `19.7.3`, a nested `dist` would just sit there in the deployed tree. Worth noting `rm -rf dist` at the top of the script means a cross-lineage duplicate is the *only* way this can trigger, so the fix is narrowly scoped.

#### Suggested patch

Making the boundary explicit rather than implied by the order of `packageNames`, so a version published in both lineages still deploys the build that actually shipped for it:

```js
// Dual lineage: 19.7.3 is the last version released as the Storybook site
// (@ovhcloud/ods-storybook); everything above it is the docs platform
// (@ovhcloud/ods-docs). The boundary is explicit rather than implied by the
// order of packageNames, so a version that ends up published in both
// lineages still deploys the build that actually shipped for it.
const packageNames = ['@ovhcloud/ods-storybook', '@ovhcloud/ods-docs'];

// Sortable key from major.minor.patch. Enough to place a version relative to
// the boundary: a prerelease never straddles it (19.7.3-alpha.0 ranks with
// 19.7.3, 20.0.0-alpha.0 ranks with 20.0.0).
const rank = (version) => version.split('-')[0].split('.')
  .reduce((acc, part) => acc * 1000 + Number(part), 0);
const STORYBOOK_LINEAGE_MAX = rank('19.7.3');
```

```js
  // Keyed by version: a version is deployed once, by the lineage that owns it.
  const lineages = new Map();
  for (const packageName of packageNames) {
    try {
      const list = JSON.parse(await $`npm view ${packageName} versions --json`);
      for (const version of list) {
        const owner = rank(version) <= STORYBOOK_LINEAGE_MAX ? packageNames[0] : packageNames[1];
        if (packageName !== owner) {
          console.error(`skipping ${packageName}@${version}: ${owner} owns this version`);
          continue;
        }
        lineages.set(version, { packageName, version });
      }
    } catch (e) {
      console.error(`no published versions for ${packageName}, skipping`, e);
    }
  }

  for (const { packageName, version } of lineages.values()) {
```

And the guard, as belt and braces once the `Map` makes it unreachable:

```js
    try {
      // Replace, never nest: same reason as the current-build block below.
      await $`rm -rf dist/v${version}`;
      await $`mv ${dir}/dist dist/v${version}`;
    } catch (e) {
```

Small bonus if you take `rank()`: it also replaces the `num()` comparator in the root `llms.txt` block, where `parseInt('0-alpha')` yields `0` and only three components are compared, so `v19.7.0` and `v19.7.0-alpha.0` tie and sort unstably. `.sort((a, b) => rank(b.slice(1)) - rank(a.slice(1)))` collapses it to one comparison, though a prerelease still ties with its release. Filtering prereleases out of that listing entirely might be the better call, since I doubt agents want to be pointed at `19.7.0-alpha.0`.

crotru:
Component creation (`pnpm new:component`) are broken for now. Do we add this in a next iteration ?

crotru:
RELEASING.md still reference storybook (next iteration too maybe?)

crotru:
In `package.json` files we still have the storybook scripts

feoche (approved):
*

crotru (approved):

PR #1111, opened 2026-08-27
feat(Input): start and end slots

crotru:
## What

Adds `startContent` and `endContent` to `Input`: decorative content placed inside
the field, before and after the value: a unit, a currency symbol, a domain
suffix, an icon, or a small ghost button. It is never part of the value and is
never submitted with the form.

## Behaviour

- Both use logical properties, so they follow the writing direction.
- The built-in action buttons (clear, mask, search) stay at the far end, after
  `endContent`, at a constant distance from the border.
- A text adornment keeps the text inset; one holding a control aligns with the
  action buttons instead.
- The `Input` never touches `aria-describedby`: decoration that carries meaning
  is silent unless the consumer references it. Documented, with an example.
- The `Input` never disables or otherwise alters the content it is given.

## Also included

- Action buttons now sit as far from the trailing border as from the top and
  bottom (2px instead of 4px), so the gap around them reads evenly.
  **This also moves Clipboard's copy button**, which insets on the same token.
- `CommandFilter` uses `startContent` for its magnifying glass, replacing the
  trailing search button.
- Fixes the cascade layer of `inputActions.module.scss` (`atoms` → `ods-atoms`).
- Hides Edge's native `::-ms-reveal`, which doubled the mask toggle.

## Testing

27 rendering e2e tests, including pixel-level geometry assertions that lock the
in-field spacing contract: there is no visual regression suite in this repo, so
they are the only guard against the spacing drifting. Plus unit tests for
`isSlotDefined` and dev stories covering RTL, constrained width, oversized
content, disabled states and in-form behaviour.

Consumer suites re-run green: Combobox, Query Filter, Command, Clipboard,
Password, Datepicker, Phone Number, Timepicker, Pagination.


packages/ods-react/src/components/command/tests/behavior/command.e2e.ts
```diff
@@ -338,6 +338,9 @@ describe('Command behaviour', () => {
 
     it('should hide a group when all its options are filtered out', async() => {
       await typeInFilter(page, 'copy');
+      // The highlight lands in an effect, one render after the list is filtered: asserting it
+      // straight after typing reads the in-between frame. Same wait as the sibling test above.
+      await page.waitForSelector('[data-ods="command-option"][aria-selected="true"]');
```
var-poro: not sure the comment is super useful

packages/ods-react/src/components/input/src/components/input-actions/inputActions.module.scss
```diff
@@ -1,15 +1,16 @@
 @use '../../../../../style/button';
 
-@layer atoms {
+@layer ods-atoms {
```
var-poro, line 3: nice catch

packages/ods-react/src/components/input/src/components/input-actions/inputActions.module.scss
```diff
@@ -1,15 +1,16 @@
...
     display: inline-flex;
+    flex: none;
     flex-flow: row;
     align-items: center;
 
     &__loading {
       display: flex;
       justify-content: center;
-      width: button.$ods-button-size-xs;
+      inline-size: button.$ods-button-size-xs;
```
var-poro: might add a comment to why this change

packages/ods-react/src/components/input/src/components/input/input.module.scss
```diff
@@ -31,23 +31,49 @@
...
+    }
+
+    &:has(> [data-ods="input-end"] :is(a, button, [role="button"])) {
+      padding-inline-end: input.$ods-input-actions-padding-right;
+    }
+
     &[data-disabled] {
       @include state.ods-is-disabled();
+
+      // Content inside an adornment keeps its own colour.
```
var-poro: doesn't feel needed

packages/ods-react/src/components/input/src/components/input/input.module.scss
```diff
@@ -31,23 +31,49 @@
...
     }
 
     &__field {
+      box-sizing: border-box;
       border: none;
       background-color: inherit;
       padding: 0;
-      width: 100%;
+      inline-size: 100%;
+      min-inline-size: 0;
```
var-poro: is this mandatory?

packages/docs/src/content/components/input/documentation.mdx
```diff
@@ -82,6 +87,14 @@ A toggle mask button can be displayed within the **Input** to toggle the visibil
 
 A search button can be displayed within the **Input** to trigger a specific action, such as performing a search.
 
+Decorative content can be placed inside the field, before the value with `startContent` and after it with `endContent` — a unit such as `kg` or `%`, a currency symbol, a domain extension, an icon, or a small ghost button. It is purely visual or interactive decoration: it is never part of the value and is never submitted with the form.
```
var-poro: Formulation makes it hard to understand : are those exhaustive cases of use, recommandations, or to do?

packages/docs/src/content/components/input/documentation.mdx
```diff
@@ -82,6 +87,14 @@ A toggle mask button can be displayed within the **Input** to toggle the visibil
...
+
+Both are placed with logical directions, so they follow the writing direction rather than a fixed side. The built-in action buttons always stay at the far end of the field, after the end content, so they remain at the same distance from the border whether or not decoration is present.
+
+The content must fit the field's content height (26px with the default theme) and must not carry a background or a border of its own. Use `size="xs"` for a `Button` — the size the built-in actions use. For a control with its own surface adjoined to the field, such as a country selector or a calendar trigger, compose elements around the **Input** instead; see **Datepicker** and **Phone Number**.
```
var-poro: do not use the default theme value in documentation as it can change

var-poro (changes requested):

feoche (approved):

packages/ods-react/src/components/input/src/components/input-actions/inputActions.module.scss
crotru: Felt like a good time to begin to use logical properties (by opportunity). You feel like it would need to be during a whole rework of this ?

packages/ods-react/src/components/command/tests/behavior/command.e2e.ts
crotru: Not even the test part actually. I remove.

packages/ods-react/src/components/input/src/components/input/input.module.scss
crotru: No. Useless since nothing would force width inside the input. I remove

packages/docs/src/content/components/input/documentation.mdx
crotru: I clarify

var-poro (approved):

PR #1112, opened 2026-08-31
WIP feat(Tag): allow rendering the tag as a link

crotru:
Tag now accepts `as`, following the convention already used by Link and BreadcrumbLink, 
so it can render a plain anchor or a routing component:

```tsx
    <Tag as="a" href="/blog/tag/kubernetes">Kubernetes</Tag>
    <Tag as={ Link } to="/blog/tag/kubernetes">Kubernetes</Tag>   // react-router
```

- the removal icon is only rendered on a button Tag: it is the removal affordance, and a link has nothing to remove
- no `type` attribute is emitted on anything but a button
- `text-decoration: none` in the tag mixin, so an anchor is not underlined

`TagProp` no longer extends the `button` attributes: those now come from whatever `as` renders. The exported signature restates the generic, so the props and the ref of the rendered element are checked at the call site (`forwardRef` alone erases the generic, which is why `<Link as="a" to="/x">` compiles today).

Tests: 5 rendering and 2 navigation e2e cases (element type, href, no icon, no type, no underline, Tab focus, Enter navigates).
Docs: link example, an "Element" variation section, and the keyboard note that only Enter activates a link Tag.

### DIFF
--- a/packages/ods-react/src/components/tag/src/components/tag/Tag.tsx
+++ b/packages/ods-react/src/components/tag/src/components/tag/Tag.tsx
@@ -1,11 +1,18 @@
 import classNames from 'classnames';
-import { type ComponentPropsWithRef, type FC, type JSX, forwardRef } from 'react';
+import { type ComponentPropsWithRef, type ElementType, type ForwardedRef, type JSX, forwardRef } from 'react';
 import { ICON_NAME, Icon, type IconName } from '../../../../icon/src';
 import { TAG_COLOR, type TagColor } from '../../constants/tag-color';
 import { TAG_SIZE, type TagSize } from '../../constants/tag-size';
 import style from './tag.module.scss';
 
-interface TagProp extends ComponentPropsWithRef<'button'> {
+interface TagProp<T extends ElementType = 'button'> {
+  /**
+   * \@default-value='button'
+   * Pass a component you may want to use as custom Tag component.
+   * Useful for example to render the tag as a link, either a plain anchor or the Link
+   * component of a routing library.
+   * */
+  as?: T,
   /**
    * @type=TAG_COLOR
    * The color preset to use.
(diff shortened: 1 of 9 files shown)


packages/ods-react/src/components/tag/src/components/tag/Tag.tsx
```diff
@@ -39,21 +51,28 @@ const Tag: FC<TagProp> = forwardRef(({
       )}
       data-ods="tag"
       ref={ ref }
-      type="button"
+      type={ isButton ? 'button' : undefined }
```
var-poro, line 54: do we really want to have type as undefined?

packages/ods-react/src/components/tag/src/components/tag/Tag.tsx
```diff
@@ -39,21 +51,28 @@ const Tag: FC<TagProp> = forwardRef(({
...
       { children }
 
       {
-        !!icon &&
+        isButton && !!icon &&
         <Icon
           className={ style['tag__close'] }
           name={ icon } />
       }
-    </button>
```
var-poro, line 52: TBD, not sure we want to automatically remove the icon if this isn't a button, what if it's not a link for example?
Also using a Link + badge might be more useful than using a tag, WYDT

packages/ods-react/src/components/tag/tests/navigation/tag.stories.tsx
```diff
@@ -13,6 +13,15 @@ export const disabled = () => (
   </Tag>
 );
 
+export const focusLink = () => (
+  <Tag
```
var-poro, line 17: Should also test other case than link

var-poro (changes requested):

packages/ods-react/src/components/tag/src/components/tag/Tag.tsx
crotru: Yes. If the tag is a `<button>` we want to have a `type="button"`, but if it is another tag we don't want a type at all

packages/ods-react/src/components/tag/src/components/tag/Tag.tsx
crotru: Ok. I went too fast and wrongly assumed icon prop was a boolean. I will come up with something else

PR #1113, opened 2026-09-02
WIP feat(Button): permit link instead of button

crotru:
Adds `as` to Button, so a navigation that needs the visual weight of a button renders a real anchor instead of a `<button>` handling a route. Prerequisite for the Pagination link mode PR.

### Notes for review
- `as` follows the existing ODS convention (Link, BreadcrumbLink, Tag). The signature is restated on the exported value because `forwardRef` erases the generic; without that, nothing about `as` is type checked.
- `ButtonProp` still extends the `<button>` props, so the seven prop types that extend it downstream are untouched. The polymorphic Omit therefore keys off a separate internal `ButtonOwnProp`, which keeps the ref following `as`.
- `_button.scss` gated all 5 state selectors on `:disabled`, which an anchor can never match, and reset no `text-decoration`. Both fixed. `ods-button()` is used only by `button.module.scss`, so the change is scoped to the component.
- Behaviour fix riding along: `disabled` is resolved once instead of being overridden by the props spread, so `<Button disabled={false} loading />` is now disabled as its prop doc claims.

### Tested
button e2e 20/20 (6 new rendering, 4 new navigation), `tsc -b` over ods-react, workspace `lint:ts`, `lint:scss`, typedoc.

Additive: minor, no MIGRATION entry.

### DIFF
--- a/packages/ods-react/src/components/button/src/components/button/Button.tsx
+++ b/packages/ods-react/src/components/button/src/components/button/Button.tsx
@@ -1,17 +1,30 @@
 import classNames from 'classnames';
-import { type ComponentPropsWithRef, type FC, type JSX, forwardRef } from 'react';
+import { type ComponentPropsWithRef, type ElementType, type ForwardedRef, type JSX, type MouseEvent, forwardRef } from 'react';
 import { SPINNER_COLOR, SPINNER_SIZE, Spinner } from '../../../../spinner/src';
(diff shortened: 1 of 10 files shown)


var-poro:
Not super fond of being able to use Button as a Link since we explicitly explain the difference between the two in the documentation. Instead we should directly use a <Link /> when it is required instead of the <Button /> component as a link IMO, open to discussion

feoche (approved):

PR #1114, opened 2026-09-02
WIP feat(Pagination): permit links instead of buttons

crotru:
Stacked on the Button `as` PR. Please target that branch.

`getPageUrl` renders every page and the previous / next triggers as real anchors, so a public listing is crawlable, bookmarkable and openable in a new tab.

### Notes for review
- The vendor already implements link mode: zag exposes `type: 'link'` + `getPageUrl({ page, pageSize })` and Ark forwards both. This is typing and context plumbing, not a reimplementation. Keeping the vendor signature keeps `pageSize` in the URL, which a pagination with a page size selector needs.
- **In link mode the URL owns the page**, and that is a correctness requirement. The component no longer advances its own state on activation: React flushes a click synchronously, so a page change made during the click rewrote the trigger `href` before the browser followed it, and "next" on page 1 landed on page 3.
  Two e2e pin it, and they click for real: a synthetic `el.click()` is not flushed the same way and hides the bug entirely.
- A dev warning fires when `getPageUrl` arrives without `page` or `defaultPage`.
  It is at render time, not in an effect, so it also fires during SSR, which is where link mode is meant to be used.
- zag never marks a trigger disabled in link mode, it only drops the `href`, so Button's anchor contract supplies `aria-disabled` and the tab order at the boundaries and when the whole pagination is disabled.
- Unchanged on purpose, and documented: the ellipsis stays a disabled button, and the go-to-page form and page size selector still report through their callbacks rather than navigating.

### Tested
pagination e2e 14/14 (7 new rendering, 5 new navigation), `tsc -b` over ods-react, workspace `lint:ts`, typedoc.

Additive: minor, no MIGRATION entry.

### DIFF
--- a/packages/ods-react/src/components/pagination/src/components/pagination-pages/PaginationPages.tsx
+++ b/packages/ods-react/src/components/pagination/src/components/pagination-pages/PaginationPages.tsx
@@ -1,4 +1,4 @@
-import { Pagination as VendorPagination } from '@ark-ui/react/pagination';
+import { Pagination as VendorPagination, usePaginationContext } from '@ark-ui/react/pagination';
 import classNames from 'classnames';
 import { type ComponentPropsWithRef, type FC, type JSX, forwardRef } from 'react';
 import { BUTTON_VARIANT, Button } from '../../../../button/src';
@@ -14,7 +14,13 @@ const PaginationPages: FC<PaginationPagesProp> = forwardRef(({
   className,
   ...props
 }, ref): JSX.Element => {
-  const { disabled, labelTooltipNext, labelTooltipPrev } = usePagination();
+  const { disabled, getPageUrl, labelTooltipNext, labelTooltipPrev } = usePagination();
+  const { nextPage, previousPage } = usePaginationContext();
+  const isLink = !!getPageUrl;
+  // In link mode zag drops the href at the boundaries but never marks the trigger disabled, as
+  // an anchor takes no `disabled` attribute. Button turns it into the `aria-disabled` contract.
+  const isPrevDisabled = isLink ? (disabled || !previousPage) : disabled;
+  const isNextDisabled = isLink ? (disabled || !nextPage) : disabled;
 
   return (
     <div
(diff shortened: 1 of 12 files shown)


packages/ods-react/src/components/pagination/src/components/pagination-pages/PaginationPages.tsx
```diff
@@ -24,7 +30,10 @@ const PaginationPages: FC<PaginationPagesProp> = forwardRef(({
       { ...props }>
       <PaginationButtonWithTooltip tooltip={ labelTooltipPrev }>
         <VendorPagination.PrevTrigger asChild>
-          <Button disabled={ disabled } variant={ BUTTON_VARIANT.ghost }>
+          <Button
```
var-poro, line 33: Not sure about this, why not use <Link /> instead?

packages/ods-react/src/components/pagination/src/contexts/usePagination.tsx
```diff
@@ -103,14 +115,23 @@ function PaginationProvider({
   const isControlled = page !== undefined;
   const currentPage = isControlled && page ? page : internalPage;
 
+  // Warned at render time, not in an effect: link mode exists for server rendered listings, and
+  // an effect never runs on the server - which is exactly where the mistake is made.
```
var-poro, line 119: 🤔

packages/storybook/stories/components/pagination/documentation.mdx
```diff
@@ -113,6 +114,38 @@ Depending on the current page number and the amount of pages, here are the diffe
...
+the page.
+
+Not everything becomes a link:
+
+- the **ellipsis** is not a page, so it stays a disabled Button with no address;
+- the **previous** trigger on the first page and the **next** trigger on the last page have no page to
+  point at, so they carry no `href` and are marked `aria-disabled`;
+- the go-to-page form and the page size selector are not navigation, so they keep reporting through
+  `onPageChange` and `onPageSizeChange` for the listing to act on.
+
```
var-poro, line 148: a bit too verbose-y, could we make this simpler?

var-poro (changes requested):

PR #1117, opened 2026-09-04
Feat(pagination): allow usage of links

crotru:
## Pagination: pages as links

Where a page is a destination of its own, the **Pagination** should render links, not Buttons. Provide `getPageUrl` and it does, styled as a `Link` rather than a `Button`: a button looks like a button, a link looks like a link.

```tsx
<Pagination
  getPageUrl={ ({ page, pageSize }) => `/products?page=${page}&size=${pageSize}` }
  page={ page }
  pageSize={ pageSize }
  totalItems={ 200 }>
  <PaginationPageSizeSelector />
  <PaginationPages />
  <PaginationPageSelector />
</Pagination>
```

**What it buys you.** Pages that can be crawled, opened in a new tab, copied and restored on reload, announced as links by a screen reader, and shipped in the server rendered HTML. The active page is not a link, since there is nowhere to go: it stays emphasised text carrying `aria-current="page"`.

**The contract.** In link mode the URL owns the state, the page and the size alike. The component renders what it is given and never moves on its own, which is what makes the whole thing safe: were it to move mid-click, it would rewrite the next href before the browser followed it and land the user one page further than the link they clicked. `onPageChange` and `onPageSizeChange` therefore fire only for the two controls that carry no destination, the "Go to page" form and the items per page Select, and the application navigates from there.

**`linkAs`** is optional. Plain anchors are enough for a server rendered site; pass your router's link component and the pages navigate inside the app instead. Triggers that lead nowhere keep no `linkAs`, as an undefined destination is meaningless to a router.

**Design.** The link hugs its digits so the underline is text width, while an invisible 40x40 overlay keeps the click target of the Button version. Same treatment for the chevrons, matching an icon only `Link`.

### Verifying

- 24 e2e tests, rendering and navigation, including the two contract cases and the hit areas.
- Two runnable applications: `packages/examples/react-router-app` (client navigation, with a counter showing the document is never reloaded) and `packages/examples/nextjs-app` (server component, links present in the initial HTML).
- Full walkthrough checked on both: page link, size change, "Go to page", out of range input, and the back button.

No change to the default Button mode.

### DIFF
--- a/packages/ods-react/src/components/pagination/src/components/pagination-item/paginationItem.module.scss
+++ b/packages/ods-react/src/components/pagination/src/components/pagination-item/paginationItem.module.scss
@@ -1,12 +1,64 @@
+@use '../../../../../style/link';
+
+// The rhythm of the whole bar, shared by the page buttons, the page links and the ellipsis.
+$ods-pagination-item-size: 40px;
+
 .pagination-item {
   display: flex;
   justify-content: center;
-  width: 40px;
+  width: $ods-pagination-item-size;
 
   &[data-selected] {
     &:not(:disabled) {
       border-color: var(--ods-theme-border-color-selected);
       background-color: var(--ods-theme-background-color-selected);
+    }
+  }
+
+  // Link mode: the item is only a box holding the rhythm of the bar, the link inside hugs its
+  // digits so that its underline never runs wider than the page number it belongs to.
+  &--link {
+    align-items: center;
+    width: auto;
+    min-width: $ods-pagination-item-size;
+    height: $ods-pagination-item-size;
+  }
+
+  // The link keeps hugging its digits, so that the Link underline stays the width of the page
+  // number. The target is carried by an invisible overlay instead, at the size of a page button:
+  // a pseudo element belongs to its anchor for hit testing, so the whole cell is clickable
+  // without the box, and therefore the underline, ever growing past the text.
+  &__link {
+    &::before {
+      position: absolute;
+      top: 50%;
+      left: 50%;
+      transform: translate(-50%, -50%);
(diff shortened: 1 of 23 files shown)


var-poro:
https://rtstatic.ovhcloud.tools/digital-tools-tech-components/ods/feat/pagination-as-links/components/pagination/examples

Links example is un-usable, is there a way to make it work?

packages/ods-react/src/components/pagination/src/components/pagination-item/paginationItem.module.scss
```diff
@@ -1,12 +1,64 @@
+@use '../../../../../style/link';
+
+// The rhythm of the whole bar, shared by the page buttons, the page links and the ellipsis.
```
var-poro, line 3: rhythm?

packages/ods-react/src/components/pagination/src/components/pagination-item/paginationItem.module.scss
```diff
@@ -1,12 +1,64 @@
...
     }
   }
+
+  // Link mode: the item is only a box holding the rhythm of the bar, the link inside hugs its
+  // digits so that its underline never runs wider than the page number it belongs to.
+  &--link {
+    align-items: center;
+    width: auto;
+    min-width: $ods-pagination-item-size;
+    height: $ods-pagination-item-size;
```
var-poro, line 24: probably would prefer to have $ods-pagination-item-width & $ods-pagination-item-height and make em point on the same value, if I recall properly that's how we do in other components

var-poro (changes requested):

PR #1118, opened 2026-09-09
feat(mcp): expose the ODS documentation to AI assistants over MCP

var-poro:
## What (PoC — draft for feedback)

New package **`@ovhcloud/ods-mcp`**: an MCP ([Model Context Protocol](https://modelcontextprotocol.io)) server that lets AI assistants (Claude Code, Cursor, VS Code Copilot, any MCP client) query the ODS documentation instead of guessing props.

```bash
claude mcp add ods -- npx -y @ovhcloud/ods-mcp
```

It is a thin read layer (~450 lines) over content we already produce and verify:

- the **committed llms set** (`packages/docs/assets/llms`) bundled into the package at build time — no scraping, no new content pipeline;
- theme tokens, recipes (with full sources) and icon names + search aliases, gathered by `scripts/copy-content.mjs`.

**Documentation version resolution** (same rule as the LLM Documentation guide): the host project's own copy first (`@ovhcloud/ods-react` ships `dist/llms` in its tarball — exact installed version, offline), then `ODS_DOCS_VERSION` against ovh.github.io, then the bundled snapshot.

**Eight tools**: `list_components`, `get_component`, `get_component_api`, `search_docs`, `get_guide`, `list_icons`, `get_tokens`, `get_recipe`.

Also adds a **"MCP Server" guide page** in the docs AI Agents section (llms-excluded like `llm-documentation`, route in the sweep).

## Integration

- Versioned in the lerna lineage (published like the other packages on the next release — the `npx` command becomes real at that point; until then, run `node packages/mcp/dist/index.js` from a checkout).
- `build:prod` / `test:spec:ci` hooks so the existing CI jobs pick the package up; **no CDS change needed**.
- `tests/run.mjs` drives the built server through the official MCP stdio client: 17 checks, including both resolution paths (project node_modules vs bundled snapshot).

## Open questions for review

- Tool naming/granularity — happy to reshape the surface before it ships.
- Should the bundled llms snapshot be trimmed (only indexes) to keep the tarball small, relying on the project/remote copies for content?
- MCP Registry publication (free, makes it one-click in VS Code) once the package is on npm.


var-poro:
## PoC field-tested — 10 scenarios through real Claude Code sessions

Ran the server through 10 non-interactive Claude Code sessions (`claude -p`, tool calls extracted from the `stream-json` event stream — not self-reported). Every scenario had all 8 tools available, so tool *routing* was exercised for real.

| Scenario | Tool calls made | Outcome |
|---|---|---|
| Datepicker full API | 1 — `get_component_api` | exact API, plus the correct insight that no prop is required and every default is internal |
| Registration form (generative) | 6 — forms guide, 3 APIs, 2 example sets **before writing code** | compiles against the real API |
| "is there an autocomplete?" | 3 — search → list → `get_component combobox` | found through the documented aliases |
| Form validation (conceptual) | 2 — `search_docs` → `get_guide forms` | ideal routing |
| Icons asked in French | 2 — filters translated spontaneously (télécharger→download) | `cog` found via alias |
| Warning color token | 1 | exact token + theme alias |
| Wrong slug ("Datagrid") | 4 — converged to `data-table` unaided | worked, but see below |
| nginx question (off-topic) | **0 ODS calls** | no routing false positive |
| `disabled` across 5 form components | 6 — self-corrected `radio` → `radio-group` | held context fine |
| **Counter-proof: same Datepicker question WITHOUT the MCP server** | — | hedged, then produced a plausible-but-wrong table: `format: 'dd/mm/yyyy'` (real prop: `dateFormatter`, a function), `disabledDays` (real: `disabledWeekDays`), invented `'en'`/`'dd/mm/yyyy'` defaults where everything is `undefined`; missed `defaultView`, `i18n`, `defaultOpen` |

The counter-proof is the pitch in one image: without the server the model *knows it is unsure* and still emits prop names that would fail to compile.

**Finding → fixed in this PR** (`f058b84c`): the "Did you mean" was substring-based, so `datagrid` suggested nothing and cost 4 round-trips. Lookups are now dash-insensitive (`datatable` resolves directly) and suggestions rank by substring / edit distance / shared stem (`datagrid` → `data-table` in one hop). Also fixed `normalizeSlug`, which lowercased before splitting camelCase — `FormField`-style slugs could never resolve. Harness grew to 20 checks, all green.

Two earlier field findings already in the branch: the pinned-version mode (`ODS_DOCS_VERSION`) now explains on stderr that it needs a docs-platform release (`llms-index.json` is not part of the Storybook-era sets online), and the guide-count check follows the content instead of hardcoding it.

var-poro:
**Exhaustiveness pass** — tested resolution ① against the *real published* `ods-react@19.7.3` tarball: it ships **no `dist/llms` at all** (the `copy:llms` embedding only exists on the unreleased platform master). The server falls through to the bundled snapshot cleanly (now pinned by a harness check, 21 total); the guide and README no longer promise installed-version resolution for pre-platform 19.x packages (`afcf85002`). Net effect today: every consumer gets the bundled snapshot until the first docs-platform release, at which point resolutions ① and ② light up on their own.

